PR #15, opened 2012-01-05
split in three projects: commons,parser and renderer

lvostinar:
The css parser is a valuable piece of code and could be used independently of the renderer part. This change clearly separates the concepts in the project. As outcome, the minimal renderer jar is replaced by three jars so that each part can be deployed separately.


stanio:
This is probably not relevant anymore as the current project is Maven based.

PR #19, opened 2012-08-05
Resolves ClassCastException for grid layouts

philcali:
## The Problem

When running `ITextRenderer.layout()` on a grid formatted html file, you would get:

```
java.lang.ClassCastException: org.xhtmlrenderer.render.BlockBox cannot be cast to org.xhtmlrenderer.newtable.TableBox
```

A great example would be to tagsoup http://twitter.github.com/bootstrap/
## The Solution

An extra check to block the exception from bubbling to the surface. This simple change resolves:
- http://code.google.com/p/flying-saucer/issues/detail?id=144

Thank you for a such a great library!

### DIFF
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/layout/BoxCollector.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/layout/BoxCollector.java
@@ -123,7 +123,7 @@ public class BoxCollector {
                 if (intersectsAggregateBounds || 
                         (container.getPaintingInfo() == null && container.intersects(c, clip))) {
                     blockContent.add(container);
-                    if (container.getStyle().isTable() && c instanceof RenderingContext) {  // HACK
+                    if (container.getStyle().isTable() && c instanceof RenderingContext && container instanceof TableBox) {  // HACK
                         TableBox table = (TableBox)container;
                         if (table.hasContentLimitContainer()) {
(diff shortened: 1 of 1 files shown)


pelodelfuego:
@Serzhan 

Hello, could you detail your workaround please ?
It would be really helpfull.

Serzhan:
@pelodelfuego You need to find 'display: table;' in each of ::after and replace them to 'display: block;'
It's work for me.

pelodelfuego:
@Serzhan 

Thanks for your reactivity,

I added:

```
*::after {
	display: block  !important;
	content: "" 	!important;
}
```

But it doesn't seems to work... I'll keep digging but maybe you have another idea ?

pelodelfuego:
solved by creating:

```
.col-print-1 {width:8%;  float:left;}
.col-print-2 {width:16%; float:left;}
.col-print-3 {width:25%; float:left;}
.col-print-4 {width:33%; float:left;}
.col-print-5 {width:42%; float:left;}
.col-print-6 {width:50%; float:left;}
.col-print-7 {width:58%; float:left;}
.col-print-8 {width:66%; float:left;}
.col-print-9 {width:75%; float:left;}
.col-print-10{width:83%; float:left;}
.col-print-11{width:92%; float:left;}
.col-print-12{width:100%; float:left;}
```

Thanks for your help !

PR #20, opened 2012-09-06
avoid extra calls to TransformerFactory.newInstance()

jerith666:
We found this to be a hot spot in profiling our application, and this change sped things up noticeably.  It didn't break anything obvious, but we would appreciate any thoughts on whether this is a safe change.  Thanks!


stanio:
I think #99 should have fixed this now?

pbrant:
That seems likely.

PR #30, opened 2013-03-05
Use display: block instead of display: table

pbrant:
Use display: block instead of display: table, display: table-row, or
display: table-section in generated content.


tran302:
Is this fix actually released?? I'm getting below error.
```
org.xhtmlrenderer.render.AnonymousBlockBox cannot be cast to org.xhtmlrenderer.newtable.TableRowBox
```

PR #36, opened 2013-08-22
Border Radius Implementation

nedjs:
Heres is full border radius support. Known Issues that should be patched
prior to a release are as follows:
- Set px value under shorthand BORDER_RADIUS is not computed correctly for DPI
- Declarations around the CSSName.BORDER_****_RADIUS should be cleaned up and re organized to be more logical and return consistant types to the BorderRadiusCorner class
- BorderRadius sides are not automatically fixed when the total radius of one side excedes its width/height. This will need to be completed after the CSS parsing issues have been resolved. (you can observe this by setting border-radius: 100%, they should be normailzed to 50% because 1 side cannot have over 100%...).
- Overflow: hidden doesnt not work properly, setClip needs to be called prior to drawing children of the Block Element.

Let me know if you have any questions or things you would like me to change with the implementation. I had a very difficult time trying to figure out how to get the CSS parsing to work the way I wanted it to but I am sure you can figure it out easily.

Note: I did  most my testing using iText 2.17, additional testing should be done with Swing and iText5

### DIFF
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/css/style/BorderRadiusCorner.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/css/style/BorderRadiusCorner.java
@@ -46,14 +46,7 @@ public class BorderRadiusCorner {
 				_leftPercent = _rightPercent = true;
 				_left = _right = value.asFloat() / 100.0f;
 			} else {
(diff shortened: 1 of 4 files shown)


Demitrius:
Hi

Not working in 9.0.7 ... ?

wsf1991:
Seems it is not working in 9.1.5 with R8, any idea?
I use html to convert to pdf. border radius does not work

PR #37, opened 2013-09-26
Added scale-to-fit feature for PDF printouts

ghost:
Added a simple boolean option to org.xhtmlrenderer.pdf.ITextRenderer that can be used to scale PDF printouts to fit the page.


dusiema:
This looks promising... What's up with the pull request?

PR #42, opened 2014-01-05
Improvement of the XHTML embedded base 64 image feature : support of CSS image embedding

vmichaud:
Improvement of the XHTML embedded base 64 image feature : support of CSS image embedding

### DIFF
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/swing/ImageResourceLoader.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/swing/ImageResourceLoader.java
@@ -65,33 +65,47 @@ public class ImageResourceLoader {
     }
 
     public static ImageResource loadImageResourceFromUri(final String uri) {
-        StreamResource sr = new StreamResource(uri);
-        InputStream is;
-        ImageResource ir = null;
-        try {
-            sr.connect();
-            is = sr.bufferedStream();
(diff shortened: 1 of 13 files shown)


flying-saucer-core/src/main/java/org/xhtmlrenderer/swing/ImageResourceLoader.java
```diff
@@ -65,33 +65,47 @@ public ImageResourceLoader(int cacheSize) {
...
 
     public static ImageResource loadImageResourceFromUri(final String uri) {
-        StreamResource sr = new StreamResource(uri);
-        InputStream is;
-        ImageResource ir = null;
-        try {
-            sr.connect();
-            is = sr.bufferedStream();
+        if (ImageUtil.isEmbeddedBase64Image(uri)) {
+            return loadEmbeddedBase64ImageResource(uri);
```
stanio: Bypassing the cache for these doesn't play nice with style sheet embedded background images used for small icons which appear hundreds, or even thousands of times in a document.  On one instance I've tried (300-400 pages) the result PDF got ~ 35 MB, while using a patched version the result PDF for the same document got ~ 2 MB.

stanio: I'll prepare a PR to improve on this.

PR #57, opened 2014-09-16
Auto scale width

anu20890:
Used Boolean variable getter/setter to auto scale width of document.


anu20890:
What is the status of this pull request? Please verify...

PR #70, opened 2015-02-05
Avoid line breaking before punctuation characters.

jensrutschmann:
Follow-up fix for pull request #60 

This introduces a new BreakIterator that delegates to BreakIterator.getLineInstance to retrieve break points. Additionally it provides breakpoints in long URLs at slash characters.

@pbrant  @tobymckoi @kmtong 
Would be nice if you could test this and let me know how it works for you.

### DIFF
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/layout/Breaker.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/layout/Breaker.java
@@ -177,7 +177,7 @@ public class Breaker {
     }
 
 	public static BreakIterator getWordStream(String s) {
-		BreakIterator i = BreakIterator.getWordInstance();
+		BreakIterator i = new UrlAwareLineBreakIterator();
 		i.setText(s);
 		return i;
 	}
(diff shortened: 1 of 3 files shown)


SvenBunge:
Please fix the issue that lines start with the punctuation of the last sentence. Due this issue we switched back to version 9.0.6.

jensrutschmann:
@SvenBunge did you try the fix from this pull request? If so, how does it work for you? Thanks

jensrutschmann:
@pbrant @tobymckoi @kmtong Has any of you had a chance to look at this? I was wondering how it works for you. We're shipping this fix in our product for some time now and it seems to work well for our customers.

I'd like to get this into fs master if there aren't any objections.

tobymckoi:
It fixes the most annoying problem with punctuation in the previous implementation, however it's technically not spec compliant (the spec is not honored by a lot of browsers anyway). The spec says line breaks should only happen on white space characters. This means 'word/////word' should not break. URLs should never break because a URL can not have a white space in it.

Testing on other browsers, Firefox mostly honors the spec except it seems to specially handle long URLs by breaking them past the 2 path. Chrome appears to only ever break on white space.

tobymckoi:
Just read the last comment from #60 about spec compliance. It appears you do get a bit of leeway with line-break behavior so it's not such a problem after all. I do like the Firefox behavior though!

afaenger:
The current behavior will break domain names and email adresses, e.g. www.xyz.com and info.de@abc.xyz.com
<pre>info.
de@abc.xyz.com</pre>

Is there any workaround to prevent such breaks in order to have a behaviour that is similar to 9.0.6 or before?

I think that white space needs to have a higher priority to fix this (as mentioned in eweiguu's comment in #60 )

pocketcap:
Is there any update on this? I am still experiencing the issue @afaenger mentioned above.
Thanks.

PR #72, opened 2015-02-19
Stream from redirected conn

tgeese:
Hi,
we have a setup were our application operates in a HTTP-context behind a proxy that has SSL enabled and that redirects HTTP to HTTPS. The HTML we pass over to flying-saucer only contains HTTP references.
The problem is that flying-saucer won't follow redirects.

We implement the handling of redirection in flying-saucer.

It would be nice if you could merge this.

### DIFF
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/swing/NaiveUserAgent.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/swing/NaiveUserAgent.java
@@ -173,9 +173,14 @@ public class NaiveUserAgent implements UserAgentCallback, DocumentListener {
 		if (needsRedirect(status)) {
 			// get redirect url from "location" header field
 			String newUrl = origin.getHeaderField("Location");
-			XRLog.load("Connection is redirected to: " + newUrl);
-			// open the new connnection again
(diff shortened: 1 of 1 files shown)


flying-saucer-core/src/main/java/org/xhtmlrenderer/swing/NaiveUserAgent.java
```diff
@@ -117,21 +120,89 @@ public void clearImageCache() {
...
+	 * This can be overwritten to customize handling of connections by type.
+	 * 
+	 * @param uri the uri to connect to
+	 * @return URLConnection opened connection to uri
+	 * @throws IOException if an I/O exception occurs.
+	 */
+	protected URLConnection openConnection(String uri) throws IOException {
+		URLConnection connection = new URL(uri).openConnection();
+		if (connection instanceof HttpURLConnection) {
+			connection = onHttpConnection((HttpURLConnection) connection);
```
stanio: May be this could be just:

```java
    ((HttpURLConnection) connection).setInstanceFollowRedirects(true);
```

-    https://docs.oracle.com/javase/8/docs/api/java/net/HttpURLConnection.html#setInstanceFollowRedirects-boolean-

PR #73, opened 2015-02-25
Same as #69 but for the itext5 version of the pdf generator

gucki:


gucki:
What about dropping support for itext < 5? Any reasons to keep it? If we keep it, we should somehow make sure that both versions don't diverge, just as they did in this case.

flyingsaucerproject:
I don't know for sure, but my guess is that a lot of people still use iText
2.x for licensing reasons.

David-seattle:
Just one data point, but my organization still uses 2.x for licensing reasons

PR #75, opened 2015-05-14
Adds a loop counter to font_face to prevent runaway threads

tdionne:
This is not pretty, but I have noticed that in some cases, when reading CSS files from a URL, this loop can enter into an infinite loop.  I have not been able to reproduce it readily, and think that it may have something to do with the socket, perhaps being reset while it is reading.  This change will at least prevent it from running away in those extreme cases.


flying-saucer-core/src/main/java/org/xhtmlrenderer/css/parser/CSSParser.java
```diff
@@ -476,7 +476,7 @@ private String medium() throws IOException {
 //      '{' S* declaration [ ';' S* declaration ]* '}' S*
 //    ;
     private void font_face(Stylesheet stylesheet) throws IOException {
-        //System.out.println("font_face()");
+//        System.out.println(font_face()");
```
tdionne: There is a bug here if the line was uncommented, oops.

PR #78, opened 2015-06-10
Fix #189 ul and ol lists don't support start index, or restart.

nfalco79:
This fix comes from patch proposed by matsonda...(full name was not accessible) on [googlecode](https://code.google.com/p/flying-saucer/issues/detail?id=189&start=100).


nfalco79:
html4 removed the ul:start and li:value attributes, but 5 re-added them.  It wouldn't hurt to support it.

Convert the following html to pdf.
<ul start="2">
  <li>blah</li>
  <li value="4">foo</li>
</ul>

What is the expected output?
2. blah
4. foo

What do you see instead?
1.blah
2.foo

PR #79, opened 2015-07-02
Add font fallback to core, not yet to change swt

willamette:
I have a project need to support font fallback, so i made this change, but only to core. Haven't done anything for SWT yet.


danfickle:
Thanks again @willamette. Font fallback is now implemented in [Open HTML to PDF](https://github.com/danfickle/openhtmltopdf) for both PDF and Java2D.

monejava:
Why isn't his change merged into mainline?

PR #80, opened 2015-07-12
Fix for Google Code Issue 231: Bold font emulation for PDF rendering causes black borders around each letter

agibagib:
As described in Google Code [Issue 231](https://code.google.com/p/flying-saucer/issues/detail?id=231), using bold font emulation causes a black border around the text.

This was happening because the stroke colour wasn't getting set prior to rendering the text.

### DIFF
--- a/flying-saucer-pdf-itext5/src/main/java/org/xhtmlrenderer/pdf/ITextOutputDevice.java
+++ b/flying-saucer-pdf-itext5/src/main/java/org/xhtmlrenderer/pdf/ITextOutputDevice.java
@@ -502,6 +502,7 @@ public class ITextOutputDevice extends AbstractOutputDevice implements OutputDev
                 float lineWidth = fontSize * 0.04f; // 4% of font size
                 cb.setLineWidth(lineWidth);
                 resetMode = true;
+                ensureStrokeColor();
             }
             if ((fontSpec.fontStyle == IdentValue.ITALIC) && (desc.getStyle() != IdentValue.ITALIC)) {
                 b = 0f;
--- a/flying-saucer-pdf/src/main/java/org/xhtmlrenderer/pdf/ITextOutputDevice.java
+++ b/flying-saucer-pdf/src/main/java/org/xhtmlrenderer/pdf/ITextOutputDevice.java
@@ -511,6 +511,7 @@ public class ITextOutputDevice extends AbstractOutputDevice implements OutputDev
                 float lineWidth = fontSize * 0.04f; // 4% of font size
                 cb.setLineWidth(lineWidth);
                 resetMode = true;
+                ensureStrokeColor();
             }
             if ((fontSpec.fontStyle == IdentValue.ITALIC) && (desc.getStyle() != IdentValue.ITALIC)) {
                 b = 0f;


jensrutschmann:
Thanks for finding this.

It seems to fix the strange font issue I have been running into in the past:
https://groups.google.com/d/msg/flying-saucer-dev/Vpf4Psz89XM/XV9kiMRsHSoJ

However it seems not only the stroke color needs to be reset, but also the line style. In my example HTML the letters have a dashed outline. With your fix the outline has the correct font color but is still dashed.

I've tried _currentPage.setLineDash(0) which seems to fix it, but I'm not sure where the correct place to invoke it is. Maybe @pdoubleya or @pbrant know more about that.

Lozic:
Suppose stroke is not needed here at all.
Just replace   
```java
cb.setTextRenderingMode(PdfContentByte.TEXT_RENDER_MODE_FILL_STROKE);
```
to
```java
cb.setTextRenderingMode(PdfContentByte.TEXT_RENDER_MODE_FILL);
```

agibagib:
@Lozic I believe the stroke is required here because it's used to emulate a bold font when a true bold font isn't available. I made the change you suggested and the text that was previously bold, isn't any more.

@jensrutschmann I'm not able to reproduce the dashed line issue you're seeing. Could you attach the html you're using so I can take a look?

jensrutschmann:
Seems it's only possible to attach images, so here's the updated HTML inline.
You can get the font file from my original email (HTML in there doesn't work anymore).
Here's a direct link:
https://groups.google.com/group/flying-saucer-dev/attach/2a1d6cc488645f5d/font-border-issue.zip?part=0.1&authuser=0

```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml">

<head>
    <title>font border issue</title>

    <style type="text/css" media="print">
        @font-face {
            font-family: LiberationSans-Regular;
            font-style: normal;
            font-weight: normal;
            src: url('file:///Users/jens/Downloads/LiberationSans-Regular-normal-normal.ttf');
            -fs-pdf-font-embed: embed;
            -fs-pdf-font-encoding: "Identity-H";
        }

        div.programlisting {
            border: 1px dashed lime;                /* Remove either this line */
        }

        body {
            font-family: LiberationSans-Regular;    /* or this line to 'fix' the issue */
        }
    </style>
</head>
<body>

                        

        
<div>

    <!--  Page 1 -->
    <div>
        <h1>Test page without code</h1>
        
        <p>some text</p>

        <div>
            <h2>More information</h2>
            <p>some text</p>
        </div>
    </div>


    <!--  Page 2 -->
    <div style="page-break-before:always;">
        <h1>Test page with code example</h1>
        
        <p>some text</p>

        <div>
            <h2>Another heading before the code</h2>
        
            <p>some text</p>

            <div class="programlisting">
                <pre>this is code</pre>
            </div>
        
            <p>some text</p>
        </div>

        <div>
            <h2>A heading after the code</h2>

            <p>some text</p>
        </div>
    </div>


    <!--  Page 3 -->
    <div style="page-break-before:always;">
        <h1>Test page without code</h1>
        
        <p>some text</p>

        <div>
            <h2>More information</h2>
            <p>some text</p>
        </div>
    </div>

</div>

</body>
</html>
```

agibagib:
@jensrutschmann I've just tested with the HTML above and I'm definitely not seeing the issue you reported in the google group. Maybe there's something else involved?

jensrutschmann:
@agibagib You're right, it's working with the latest code from fs master. I had your change applied to an earlier version that didn't include some changes made to the font resolver in the meantime.
So just forget what I said :-)

nfalco79:
Thanks very much for this patch, we need so much this patch.
Possible that nothing of the fs team can do nothing abiut this (and some others patches)?!?

PR #81, opened 2015-07-15
license update to LGPLv3

schauerd:
- license update LGPL version 2.1 to 3
- added license text of GPLv3 (mandatory)
- links updated
- added license information in poms
- added license includes for jars


LICENSE
```diff
@@ -31,13 +32,13 @@ License: Apache Software License Version 2.0
 Not packaged with release; development using version 1.6.x
 
 iText (PDF generation)
-http://www.lowagie.com/iText/ and http://itextpdf.sourceforge.net/
+http://itextpdf.com/
 License: Mozilla Public License Version 1.1
-Using version 2.0.8.
-Included as lib/iText-2.0.8.jar
+Using version 2.1.7.
+Included as lib/iText-2.1.7.jar
```
jensrutschmann: I think we should also mention itext5 here, which is available under a different license (AGPL).

flyingsaucerproject:
Oh, this isn't quite what I was thinking. I'm fine with LGPL 2.1 or later
(and of course including the license text in the JARs etc). I'm not fine
with just LGPL 3. For example, see
http://nmav.gnutls.org/2013/03/the-perils-of-lgplv3.html?m=1

For me the goal is that FS is used as widely as possible. LGPL 2.1 or later
does that more effectively.

Sorry for any confusion. Could you resubmit?

schauerd:
OK. No problem. I will create a new pull request.

PR #83, opened 2015-07-22
Fix integer arithmetic rounding error in BorderPainter.generateBorderShape()

dmap:
This error causes a (incorrect) 1 pixel offset when rendering borders for boxes that have either a width or height that is an odd number.

The error occurs in the final translation of the border into position.  Seems like it was a simple oversight given that the same error is avoided two lines above. 

PS:  Any chance of a 9.0.8 release any time soon, would be nice to be able to pull this version from maven central...


dmap:
Added two more fixes to various text width calculations.  These caused layout issues as described in the commit comments.

dmap:
Excellent, thanks for the merge.

PR #84, opened 2015-09-03
Fixed build on UNIX-like operating systems.

jensrutschmann:
Specifying "/META-INF" as target for the license files in the <resources> tag will result in maven trying to put the license files into a directory "META-INF" located directly under the filesystem root. This directory does not exist and is outside of the maven build folder. Therefore we need to prefix the target directory with the maven property "${project.build.outputDirectory}" which points to the respective "target/classes" subfolder in each maven module.


jensrutschmann:
@schauerd Please verify that this works for you as well.

PR #91, opened 2016-02-24
Fix for wrapping on div as seen on chrome and firefox

avinash10584:
There an issue with the last commit in Breaker.java , most browsers at end of the line trim the spaces in block, so for an example if I have text "A Certificate will be issued in view of:" and the block can only have content of length "A Certificate will be issued" (notice missing space in end) then the chrome and firefox are keeping the word 'issued' on same line and breaking after that,

image 5
image 6

Chrome and firefox are behaving close to second scenario and are allowing the text even if it has trailng spaces going out of bounds of block. The old code without UrlAware breaker was behaving close to the chrome and firefox

```
<div style="width:675px;background-color:orange;font-size:f10pt;font-family:'Arial'">
        Prosecution on the merits is (or remains) closed in this inter partes reexamination proceeding.  This proceeding is subject to reopening at the initiative of the Office or upon petition.  <span class="italics">Cf</span>. 37 CFR 1.313(a).  A Certificate will be issued in view of:
      </div>
```

I have added the fix by allowing extra calculation on graphics length when it is reaching end of line

### DIFF
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/layout/Breaker.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/layout/Breaker.java
@@ -115,11 +115,15 @@ public class Breaker {
         int lastWrap = 0;
         int graphicsLength = 0;
         int lastGraphicsLength = 0;
-
+        int lastTokenLength = 0;
+        String currentToken = "";
+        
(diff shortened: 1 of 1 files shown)


pbrant:
@jensrutschmann Jens, could you take a look at this? (another line breaking fix)

PR #92, opened 2016-03-10
Border radius with zero-width border fix

fenrhil:
Defining a border-radius and a zero-width border on a box allows to make the background of this box "round" on the corners.
Currently, if we have a box with border-radius, but the border has a zero-width, then we get a `ClassCastException` in `java.awt.geom.Area.intersect(Area) ` due to some paths being invalid.

Here are a JUnit test case (which fails before the fix) **and a fix.**
Could you please merge this and release a new version?

This isssue is rather blocking to my team, since we use the same CSS for both our web application and PDF rendering (with a few additions). Thus the new "border-radius" feature completely broke our PDF-generation, since there already is some CSS with "border-radius".

### DIFF
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/render/BorderPainter.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/render/BorderPainter.java
@@ -20,19 +20,15 @@
 package org.xhtmlrenderer.render;
 
 import java.awt.BasicStroke;
-import java.awt.Polygon;
 import java.awt.Rectangle;
-import java.awt.RenderingHints;
 import java.awt.Shape;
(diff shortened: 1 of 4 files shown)


pbrant:
Thanks. I'll push out a new release this weekend.

fenrhil:
Hello @pbrant ,

It seems there's been no release, after all. Would it be possible to push out a new release ?

Thanks,

pbrant:
You're right. My apologies. This is done. It's on bintray now
https://bintray.com/flyingsaucerproject/maven/ and should be on Central in
a couple of hours.

fenrhil:
Thank you. I confirm the issue was fixed with this new release.

PR #93, opened 2016-04-12
@font-face bug

maczam:
I committed a test to reproduce this bug。


synkarius:
@maczam I know this is super-old, but if you're still on Github and you still remember, what was this bug?

asolntsev:
@maczam I will merge your PR, but could you please recall what is the problem with @font-face?
Your PR generates an image, but what's wrong with this image?

PR #98, opened 2016-09-30
Disable (external) entity processing

lillesand:
Currently all uses of flyingsaucer that doesn't use a configured parser, which I assume are quite a few, will be vulnerable to [XXE attacks](https://www.owasp.org/index.php/XML_External_Entity_(XXE)_Processing) (for some simple test cases, see [https://www.gracefulsecurity.com/xxe-cheatsheet/](grateful security)). Applications can fix this using configured parser, but insecure by default usually just means insecure.

Disallowing doctype declarations altogether probably won't fly for all existing use cases for flyingsaucer, but I want show all options to spark discussion. I see at least two viable options for this:

1. Disallow all entity expansion by default, but make it configurable. This will break backwards compability for all upgrading parties which can't configure the parser.
2. Try to disable external entities specifically. This can cause some hit'n'miss depending on which parser implementations are used.

Thoughts?

### DIFF
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/resource/XMLResource.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/resource/XMLResource.java
@@ -23,6 +23,7 @@ import java.io.InputStream;
 import java.io.Reader;
 import java.util.logging.Level;
 
+import javax.xml.XMLConstants;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.SAXParser;
 import javax.xml.parsers.SAXParserFactory;
@@ -44,6 +45,7 @@ import org.xml.sax.SAXNotSupportedException;
 import org.xml.sax.SAXParseException;
 import org.xml.sax.XMLReader;
 import org.xml.sax.helpers.XMLReaderFactory;
+import java.util.logging.Level;
 
 
 /**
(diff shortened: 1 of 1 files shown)


lillesand:
@pbrant do you still do work on flyingsaucer?

danfickle:
I just released RC7 of the [Open HTML to PDF](https://github.com/danfickle/openhtmltopdf) fork which should fix this vulnerability. Thanks again to @lillesand

stanio:
> This is probably overdoing it a bit, but hopefully it will spark a discussion about what is the correct level of entity support for flyingsaucer.

Yes, this seems a bit unnecessary.  All it should need is:
```diff
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/resource/FSEntityResolver.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/resource/FSEntityResolver.java
@@ -28,6 +28,7 @@ import org.xml.sax.SAXException;

 import java.io.IOException;
 import java.io.InputStream;
+import java.io.StringReader;
 import java.net.URL;
 import java.util.HashMap;
 import java.util.Map;
@@ -58,6 +59,18 @@ public class FSEntityResolver implements EntityResolver {
      */
     private static FSEntityResolver instance;

+    private static InputSource emptyEntity = new InputSource(new StringReader("") {
+        public int read() {
+            return -1;
+        }
+        public int read(char[] cbuf, int off, int len) {
+            return -1;
+        }
+        public void close() {
+            // no effect
+        };
+    });
+
     private final Map entities = new HashMap();

     // fill the list of URLs
@@ -122,9 +135,9 @@ public class FSEntityResolver implements EntityResolver {
             XRLog.xmlEntities(Level.FINE, "Entity public: " + publicID + " -> " + url +
                     (local == null ? ", NOT FOUND" : " (local)"));
         } else {
-            XRLog.xmlEntities("Entity public: " + publicID + ", no local mapping. Parser will probably pull from network.");
+            XRLog.xmlEntities("Entity public: " + publicID + ", no local mapping. Replacing with an empty content.");
         }
-        return local;
+        return (local == null) ? emptyEntity : local;
     }

     /**

```

Even if one doesn't want to mess with the `FSEntityResolver` implementation as currently is, one could set it up in `XMLResource.XMLResourceBuilder.addHandlers(XMLReader)` like:

```java
                // add our own entity resolver
                xmlReader.setEntityResolver(new EntityResolver() {
                    private InputSource empty = new InputSource(new StringReader("") {
                        public int read() {
                            return -1;
                        }
                        public int read(char[] cbuf, int off, int len) {
                            return -1;
                        }
                        public void close() {
                            // no effect
                        };
                    });
                    public InputSource resolveEntity(String publicId, String systemId) throws SAXException, IOException {
                        InputSource entity = FSEntityResolver.instance().resolveEntity(publicId, systemId);
                        return (entity == null) ? empty : entity;
                    }
                });
```

On another note, most of the `XMLResource` looks like a complete mess:

-   A SAX parser to feed an identity transformer just to produce a DOM `Document`; moreover
-   Creating a `DocumentBuilderFactory` to obtain a `DocumentBuilder` just to create an empty `Document` for the result of the transformation – the transformation will create a `Document` anyway;
-   All these parser and transformer factories instantiated every time – quite ineffective.

I'll prepare a PR to get some feedback on it.

pbrant:
@stanio Thanks much for taking a look at this. A PR would be very much appreciated.

PR #99, opened 2016-11-15
Clean XMLResource up

stanio:
Related to my comment to #98, here are my suggestions.

Have a look at the first two:

-   Ensure no external entities get loaded
-   ... or set it up already in FSEntityResolver (?)

If you're fine with the later – I'll just squash the two, otherwise I'll rewrite to leave just the first.

The cleanup part is probably rather obvious.

I'm open for suggestions regarding the last one:

-   Cache and reuse parser instances

I've given alternative approach in the commit message for that one.

### DIFF
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/resource/FSEntityResolver.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/resource/FSEntityResolver.java
@@ -28,6 +28,7 @@ import org.xml.sax.SAXException;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.StringReader;
 import java.net.URL;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


stanio:
The overall diff may be seen a bit better [ignoring the whitespace changes](99/files?w=1).

stanio:
> Cache and reuse parser instances

Haven't done similar for `createXMLResource(javax.xml.transform.Source)` as I think it has little to no, or just less utility than the other.

pbrant:
This all looks good to me (including the ThreadLocal cache). If you want to squash or otherwise rearrange the commits involved, that's fine with me.

Otherwise, I'm happy to merge as is. Do you want these changes released once they're in?

stanio:
On a second thought, I don't want to squash anything at this point.

I'm not in a hurry for having these changes in an official release – whenever you think is appropriate, or if there's a demand for it.

pbrant:
Thanks.

PR #101, opened 2016-11-18
[add_osgi_support] Adding OSGi support

trohrberg:
Adding OSGi Support to Flying Saucer

The goal of these changes were to enable the consumption of the flying saucer libraries within an OSGi context. For having full information about dependencies, OSGi needs some additional meta-data in the MANIFEST.MF files (i.e. Bundle-SymbolicName, Export-Package, Import-Package, etc.). With that meta-data in place, one can but the Maven artifacts into a p2 repository (e.g. via Package Drone) and consume the flying saucer libraries in a Maven/Tycho build for example.

For almost all artifacts - apart from the flying-saucer-pdf one - adding the required meta-data could be achieved by adding the maven-bundle-plugin to the build execution in the pom.xml. With a little tweaking this bundle generates the required meta-data into the MANIFEST.MF files without touching the existing build configuration.

However, for the flying-saucer-pdf artifact this approach did not work since that one declares a dependency to a very old version of itest due to licensing requirements. This dependency would have to be reflected as an Import-Package statement which would require the itext library of that very old version also OSGi-aware. Since the latter was not possible to be done, another approach has been chosen: A new artifact flying-saucer-pdf-osgi has been created that re-bundles the original flying-saucer-pdf artifact as an OSGi bundle. OSGi users just use that new artifact instead of the original one and in fact are consuming the same binary result as if they would consume the flying-saucer-pdf artifact directly. All Maven users can still use the original flying-saucer-pdf artifact.

Remark: I'm not completely sure about the details of the tweaking of the maven-bundle-plugin in the individual pom.xml files. However, the current state works for us in a product building on top of Eclipse Equinox and built with Maven/Tycho. I'm of course open to discussions about changing that configuration. Please feel free to push any changes you think are reasonable or get in touch with me via discussion.

### DIFF
--- a/flying-saucer-core/pom.xml
+++ b/flying-saucer-core/pom.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
 
   <modelVersion>4.0.0</modelVersion>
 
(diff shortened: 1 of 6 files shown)


pbrant:
Thanks much.

trohrberg:
Hey Peter,

thank you very much for merging my pull request. Now that those changes are in the master, it would be totally great for us if there was a new release containing those changes as quickly as possible. Then we could use the "official" version again over our custom fork.

Regards
Timo

PR #103, opened 2016-11-20
Create bookmarks from HTML outline automatically

stanio:
1. Doesn't need non-standard elements in the `<head>` which most likely duplicate the HTML outline
2. Doesn't require explicit anchors present
3. Simple usage – automatic, and allows for fine customization with `data-pdf-bookmark` attribute hints

When bookmarks have been already defined using `<bookmarks>` – no result change.  To suppress the automatic bookmarks generation completely one could specify:

```html
<html data-pdf-bookmark="none">...
```

### DIFF
--- a/flying-saucer-pdf/src/main/java/org/xhtmlrenderer/pdf/HTMLOutline.java
+++ b/flying-saucer-pdf/src/main/java/org/xhtmlrenderer/pdf/HTMLOutline.java
@@ -1,0 +1,145 @@
+package org.xhtmlrenderer.pdf;
+
+import java.util.ArrayList;
+import java.util.IdentityHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.w3c.dom.Element;
+import org.w3c.dom.Node;
+import org.w3c.dom.traversal.DocumentTraversal;
+import org.w3c.dom.traversal.NodeFilter;
+import org.w3c.dom.traversal.NodeIterator;
+
+import org.xhtmlrenderer.pdf.ITextOutputDevice.Bookmark;
+import org.xhtmlrenderer.render.Box;
+
+class HTMLOutline {
+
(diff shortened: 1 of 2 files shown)


stanio:
See also 00e76d9b3530d1dff1d33c9201dc858ecac3bf5b

I guess this PR needs to include the same "pdf-itext5" changes, also.  The new source file likely needs a file header including the license.  I'll gladly address all of this once I'm given a preliminary feedback on the functional change I'm suggesting.

stanio:
As a demo one could try this rather tough document:

https://www.w3.org/TR/xmlschema-1/

stanio:
As a side note, I've first thought this could/should be best controlled via style sheet extensions, like (possible user-agent styles):

```css
h1 { -pdf-bookmark: 1 }
h2 { -pdf-bookmark: 2 }
h3 { -pdf-bookmark: 3 }
h4 { -pdf-bookmark: 4 }
h5 { -pdf-bookmark: 5 }
h6 { -pdf-bookmark: 6 }

/* Exclude outline of nested sectioning roots. */
blockquote :any(h1, h2, h3, h4, h5, h6),
td :any(h1, h2, h3, h4, h5, h6) {
  -pdf-bookmark: none;
}
```

(See [`:any`](https://developer.mozilla.org/en-US/docs/Web/CSS/:any))

Possible user styles:

```css
/* Exclude individual heading from outline. */
h3.extra { -pdf-bookmark: none }

/* Establish non-heading bookmark. */
tr.extra {
  -pdf-bookmark: 4 attr(data-bookmark-name);
}
```

The last one would mean establish an implicit section of level/rank 4, and set the bookmark name to the value of the "data-bookmark-name" attribute of the element (vs. using the element [text content](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent)).

All this require more thought on how the given property cascades, etc., and I don't currently have enough knowledge of the CSS implementation in `flyingsaucer`.

pbrant:
Looks good! Thanks. This is a much better approach than before.

PR #104, opened 2016-11-22
PDF bookmarks from HTML outline – clean and polish

stanio:
Amends #103

-   Use separate values for "none" (no) bookmark but consider children, and "exclude" tree from outline;
-   Expand a bit on documentation;
-   Add automatic HTML outline with the iText 5 library, also.

### DIFF
--- a/flying-saucer-pdf-itext5/src/main/java/org/xhtmlrenderer/pdf/HTMLOutline.java
+++ b/flying-saucer-pdf-itext5/src/main/java/org/xhtmlrenderer/pdf/HTMLOutline.java
@@ -1,0 +1,235 @@
+/*
+ * {{{ header & license
+ * Copyright (c) 2016 Stanimir Stamenkov
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public License
+ * as published by the Free Software Foundation; either version 2.1
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+ * }}}
+ */
+package org.xhtmlrenderer.pdf;
+
+import java.util.IdentityHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.w3c.dom.Element;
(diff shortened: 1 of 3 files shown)


pbrant:
Thanks!

jgelo:
Is there a way to prevent the outline from being generated using a style or data attribute? 

`<html data-pdf-bookmark="none">` had no effect and I'd rather not alter the java source.

pbrant:
@stanio Do you have any other changes you'd like to make? If not, I'll release a new version from master this weekend per @trohrberg's request.

stanio:
@jgelo, no style support currently.  If you take a look at the javadoc for `HTMLOutline.generate()`:

> #### Example document customizations
>
> ##### Include non-heading element as bookmark (level 4)
>
>     <strong data-pdf-bookmark="4">Foo bar</strong>
>
> ##### Specify bookmark name
>
>     <tr data-pdf-bookmark="5" data-pdf-bookmark-name="Bar baz">...</tr>
>
> ##### Exclude individual heading from bookmarks
>
>     <h3 data-pdf-bookmark="none">Baz qux</h3>
>
> ##### Prevent automatic bookmarks for the whole of the document
>
>     <html data-pdf-bookmark="exclude">...</html>

stanio:
@pbrant, I don't have additional changes currently.  I would like to see a new version including the changes so far, too.  Thank you.

stanio:
> ... take a look at the javadoc for `HTMLOutline.generate()`...

To clarify, automatic (and customized via `data-pdf-bookmark` attributes) HTML outline doesn't take place if `<bookmark>` elements have been given in the HTML `<head>`:

```html
<html>
<head>
  ...
  <bookmarks>
    <bookmark ... />
    <bookmark ...>
      <bookmark ... />
    </bookmark>
  </bookmarks>
</head>
<body>
  ...
</body>
</html>
```

This preserves the current behavior related to usage of `<bookmark>` elements.  What has changed is documents which don't use `<bookmark>` elements now get automatic bookmarks according to the outline produced from the headings found in the document.  One may opt-out of this new behavior by specifying `data-pdf-bookmark="exclude"` on the root `<html>` element, for example.  Otherwise the result bookmarks may be customized further, as per the examples given.  One example I haven't given explicitly is:

> ##### Include the outline of nested sectioning root
>
>     <blockquote data-pdf-bookmark="none">
>       <h4>...</h4>
>       ...
>     </blockquote>

By default the outline of such nested [sectioning roots](https://www.w3.org/TR/html51/sections.html#sectioning-roots) is excluded (`data-pdf-bookmark="exclude"`) from bookmarks.  Specifying `data-pdf-bookmark="none"` on them would cause them not generating a bookmark, but considering their content/nested outline to be included in the overall document outline.

jgelo:
@stanio, thank you for the info and examples, very much appreciated!  I'm mainly using saucer to generate reports where outline/bookmarks wouldn't make sense so I prefer to opt out entirely, however, I do have other content where outline/bookmarks would be beneficial to include.

pbrant:
@stanio @trohrberg https://github.com/flyingsaucerproject/flyingsaucer/releases/tag/9.1.1 was released over the weekend

PR #105, opened 2016-12-03
Enable parsing HTML5 doctype documents which use HTML entities

stanio:
Enable successful parsing of documents like:

```xml
<!DOCTYPE html>
<html>
<head>
  <title>HTML entities test</title>
</head>
<body>

<h1>Test:&ensp;HTML&nbsp;entities</h1>

</body>
</html>
```

or:

```xml
<!DOCTYPE html SYSTEM "about:legacy-compat">
<html>
<head>
  <title>HTML entities test</title>
</head>
<body>

<h1>Test:&ensp;HTML&nbsp;entities</h1>

</body>
</html>
```

See [_§ 8.1.1 The DOCTYPE_](https://www.w3.org/TR/html5/syntax.html#doctype-legacy-string) of the HTML5 specification for details.

This makes use of the SAX2 [`EntityResolver2.html#getExternalSubset()`](http://www.saxproject.org/apidoc/org/xml/sax/ext/EntityResolver2.html#getExternalSubset%28java.lang.String,%20java.lang.String%29) interface.

**Parser implementation notes**

Using genuine Xerces implementation this allows parsing no `DOCTYPE` documents which use HTML entities, also:

```xml
<html>
<head>
  <title>HTML entities test</title>
</head>
<body>

<h1>Test:&ensp;HTML&nbsp;entities</h1>

</body>
</html>
```

The root document element has to be `html` (case-insensitive).

The Xerces fork included in the Oracle Java SE distribution doesn't seem to fire `getExternalSubset()` when no `DOCTYPE` present, although the SAX specification says it should.

*DTD internal subset*

Using the Oracle Java SE bundled JAXP implementation, specifying a DTD internal subset on a document:

```xml
<!DOCTYPE html [
  <!ENTITY nbsp "&#160;">
  <!ENTITY ensp "&#8194;">
]>
<html>
<head>
  <title>HTML entities test</title>
</head>
<body>

<h1>Test:&ensp;HTML&nbsp;entities</h1>

</body>
</html>
```

in combination with `getExternalSubset()` returning non-`null` source seems to cause all text nodes missing from the result document.  If you need to parse such documents either plug official Xerces jar on the classpath, or build `org.w3c.dom.Document` instances using own parser configuration.

Note, this bug doesn't appear to happen if one also specifies an external subset:

```xml
<!DOCTYPE html SYSTEM "urn:x-dummy" [
  <!ENTITY nbsp "&#160;">
  <!ENTITY ensp "&#8194;">
]>
<html>
<head>
  <title>HTML entities test</title>
</head>
<body>

<h1>Test:&ensp;HTML&nbsp;entities</h1>

</body>
</html>
```

In this case no `getExternalSubset()` is called but `resolveEntity()`.


stanio:
On a separate note, looks like Flying Saucer bundles the HTML 4 DTDs which use SGML syntax which can't be read by an XML parser.  These are configured to use in/with `FSEntityResolver`.  I wonder if it makes sense to resolve the HTML 4 public identifiers to the corresponding XHTML 1.0 DTDs, instead.  Both should include the same named entity declarations but ultimately, as Flying Saucer doesn't perform XML validation, all it needs are the named entity declarations, and as the [HTML5 specification suggests](https://www.w3.org/TR/html5/the-xhtml-syntax.html#parsing-xhtml-documents):

> This specification provides the following additional information that user agents should use when retrieving an external entity: the public identifiers given in the following list all correspond to [the URL given by this link](https://w3c.github.io/html/entities.dtd). (This URL is a DTD containing the entity declarations for the names listed in the [named character references](https://www.w3.org/TR/html5/syntax.html#named-character-references) section.) [XML]
>
> *   `-//W3C//DTD XHTML 1.0 Transitional//EN`
> *   `-//W3C//DTD XHTML 1.1//EN`
> *   `-//W3C//DTD XHTML 1.0 Strict//EN`
> *   `-//W3C//DTD XHTML 1.0 Frameset//EN`
> *   `-//W3C//DTD XHTML Basic 1.0//EN`
> *   `-//W3C//DTD XHTML 1.1 plus MathML 2.0//EN`
> *   `-//W3C//DTD XHTML 1.1 plus MathML 2.0 plus SVG 1.1//EN`
> *   `-//W3C//DTD MathML 2.0//EN`
> *   `-//WAPFORUM//DTD XHTML Mobile 1.0//EN`
>
> Furthermore, user agents should attempt to retrieve the above external entity's content when one of the above public identifiers is used, and should not attempt to retrieve any other external entity's content.

That is, all of these (optionally including the HTML 4 public identifiers) should resolve to the HTML5 DTD.

PR #107, opened 2016-12-04
Replace ThreadLocal cache with simple object pool

stanio:
Borrows few internal implementation changes from #106 without introducing user API complications:

-   Allow recycling parsers between multiple threads
-   Allow configuring `ignoringElementContentWhitespace` of the parser factory

The last one employs probably a little known fact Xerces (including its fork included in the Oracle Java SE distribution) has ever been able to report [_"ignorable whitespace"_](http://www.saxproject.org/apidoc/org/xml/sax/ContentHandler.html#ignorableWhitespace%28char[],%20int,%20int%29), even when not validating (contrary to what the [`DocumentBuilderFactory` API doc](https://docs.oracle.com/javase/7/docs/api/javax/xml/parsers/DocumentBuilderFactory.html#setIgnoringElementContentWhitespace%28boolean%29) suggests), and even with just partial grammar provided.

The `ignoringElementContentWhitespace = true` configuration enables significant memory savings for table-data heavy documents, for example.

---

New configuration properties introduced:

-    `xr.load.parser-pool-capacity` (default: `3`)
-    `xr.load.ignore-element-content-whitespace` (default: `false`)


pbrant:
Awesome, thanks.

PR #108, opened 2016-12-10
Give XMLResrouce.load(javax.xml.transform.Source) some credit

stanio:
-   Provide basic "parser" caching with `XMLResrouce.load(javax.xml.transform.Source)` as for `XMLResource.load(org.xml.sax.InputSource)`
-   Document example usage of `XMLResrouce.load(javax.xml.transform.Source)` for parsing real-world HTML

### DIFF
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/resource/XMLResource.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/resource/XMLResource.java
@@ -162,51 +162,20 @@ public class XMLResource extends AbstractResource {
 
     private static class XMLResourceBuilder {
 
-        private final Queue<Reference<DocumentBuilder>> parserPool =
-                new ArrayBlockingQueue<Reference<DocumentBuilder>>(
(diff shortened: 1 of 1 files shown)


stanio:
In theory, we could route all `XMLResource.load(InputSource)` requests like:

```java
    public static XMLResource load(InputSource source) {
         XMLReader xmlReader = XMLReaderFactory.createXMLReader();
         return load(new SAXSource(xmlReader, source));
    }
```

which would allow to configure only `org.xml.sax.driver` system property to plug in the TagSoup parser globally, but then the underlying identity transform used to produce a `org.w3c.dom.Document` is about x5 slower than `DocumentBuilder.parse()`, even when caching/reusing the `XMLReader` instances.

Such implementation has been in place before my clean up in 0e5408111fc08eb73e2cecc90c86fb87731496d7.  I may restore the behavior conditionally testing if `xr.load.xml-reader` is set.

stanio:
> Such implementation has been in place before my clean up in 0e5408111fc08eb73e2cecc90c86fb87731496d7.  I may restore the behavior conditionally testing if `xr.load.xml-reader` is set.

I've completely restored the behavior in #109.

PR #110, opened 2017-01-16
Add entity reference support

cbxp:
fixed rendering of `&` and other html entities


pbrant:
Thanks. Let me know if you'd like a release with this change.

PR #111, opened 2017-01-21
Implemented css3 word-wrap capability

LZaruba:
Hi,
I have redesigned Breaker to allow usage of pluggable text breakers. Default behaviour mostly copies original state. There is difference in the Breaker though - original code was trying to squeeze text by removing trailing spaces. New code doesn't do that anymore.

To leverage described modularity I have created new module flying-saucer-fop which implements text breaker using Apache FOP library with included word wrapping dictionaries from project OFFO. I am bit unsure about licence compatibility so please verify that. flying-saucer-fop module contains test that shows usage including language definition. I will be happy to provide snippets for the project documentation if needed.

As a step in the process of solving my issue with text layout I came across spaces being left at the end of the line after wrapping which caused text-align: justify to work poorly. I've fixed that in [InlineBoxing](https://github.com/LZaruba/flyingsaucer/blob/master.word-wrap/flying-saucer-core/src/main/java/org/xhtmlrenderer/layout/InlineBoxing.java#L241-L243) but I am not really sure about the solution - if this is the right place and what about text-align: right and maybe others. I can strip trailing spaces inside of the Breaker. Anyway it works now as you can see in the example.

Feel free to contact me for further improvements and information as this is the first version.

Thanks for the great library!
Cheers
LZ

### DIFF
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/css/constants/CSSName.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/css/constants/CSSName.java
@@ -1070,6 +1070,18 @@ public final class CSSName implements Comparable {
                     INHERITS,
                     new PrimitivePropertyBuilders.WordWrap()
             );
+    
+    /**
+     * Unique CSSName instance for CSS3 property.
+     */
+    public final static CSSName HYPHENS =
+            addProperty(
+                    "hyphens",
+                    PRIMITIVE,
+                    "none",
+                    INHERITS,
+                    new PrimitivePropertyBuilders.Hyphens()
+            );
 
     /**
      * Unique CSSName instance for CSS2 property.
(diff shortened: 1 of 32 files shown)


pbrant:
Wow, this is super cool. I'll take a look tomorrow.

pbrant:
Looks good to me. Thanks! Let me know if/when you'd like a release with these changes.

I'm no lawyer, but I don't think using code derived from code under an Apache 2 license should be a problem. In any case, the heritage is clear.

LZaruba:
Hi,

Ok, thanks. I am currently using the fork in testing so we will see in a week if there is something not ready for production.

I guess that I will also need to add some kind of weights to the break points to prevent accessive usage of hyphenation. We could probably wait for that as well.

I will keep you posted.

Have a nice day.

LZ

Lukas Zaruba

PR #113, opened 2017-01-30
Add Maven profiles for choosing between OpenPDF or iText as dependency

andreasrosdal:
Add Maven profiles for choosing between OpenPDF or iText as dependency.

OpenPDF is a Java library for creating and editing PDF files with a LGPL and MPL open source license. OpenPDF is based on a fork of iText 4.  https://github.com/LibrePDF/OpenPDF

iText has a AGPL license, but is sold as a commercial product.  http://itextpdf.com/


pbrant:
Thanks!

PR #116, opened 2017-03-08
Resolve uri regression fix

huxi:
This PR fixes the regression described in https://code.google.com/archive/p/flying-saucer/issues/262

`baseURI` handling for URIs like `jar:file:/path/jarfile.jar!/foo/index.xhtml` used to work in flyingsaucer <= 9.0.6 but stopped working with 9.0.7.

On a side note:
What is the proper way to get in touch with the people working on flyingsaucer? I can't find an active issue tracker anywhere.
- https://xhtmlrenderer.java.net/ is outdated.
  - The mailinglists at https://java.net/projects/xhtmlrenderer/lists/users/archive and https://java.net/projects/xhtmlrenderer/lists/dev/archive seem to be inactive.
- https://code.google.com/archive/p/flying-saucer/ is archived but the only link given in `README.md`.
- The issue tracker at https://code.google.com/archive/p/flying-saucer/issues is also archived.
- https://github.com/flyingsaucerproject/flyingsaucer/ does only allow pull requests but not issues.
- https://flyingsaucerproject.github.io/flyingsaucer/ says "test page".
- https://flyingsaucerproject.github.io/ is entirely empty.

I think I just found out that https://groups.google.com/forum/#!forum/flying-saucer-users would be the correct place but that info was hidden *really* good.

### DIFF
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/swing/NaiveUserAgent.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/swing/NaiveUserAgent.java
@@ -352,7 +352,7 @@ public class NaiveUserAgent implements UserAgentCallback, DocumentListener {
      */
     public String resolveURI(String uri) {
         if (uri == null) return null;
-        String ret = null;
+
         if (_baseURL == null) {//first try to set a base URL
             try {
                 URI result = new URI(uri);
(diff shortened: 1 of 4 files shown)


pbrant:
Thanks much. I'll do a release today. If you could make a pass through README.md to get rid of obsolete or misleading info, I'd happily merge it.

The old java.net and Google Code are indeed dead. I didn't realize that github issues were disabled. I don't mind enabling them, but at this point the only changes will be through PRs so it might be just as well to leave it disabled.

huxi:
I created #118.

Thanks a lot for the swift reaction!

pbrant:
I've released master as https://github.com/flyingsaucerproject/flyingsaucer/releases/tag/v9.1.2 . It's on Bintray now and should sync to Central in a couple of hours.

PR #127, opened 2017-08-03
Fix infinite loop when using float and word-wrap:word

liias:
This fixes two issues:
1) Calculating where word-wrapped content should go.
2) Relying on do-while loop having successful end condition - now limiting iterations to 100 000 for inline element parsing.

Discussion here: https://groups.google.com/forum/#!topic/flying-saucer-users/0b98p88ZxQc

I made an educated guess with the fix, not really sure how the existing codebase should work, so please review the change.


pbrant:
Thanks!

PR #128, opened 2017-08-24
Getting HTML data from bytes array and returning resuls as the bytes array

aksndr:
Modifications allows to embed library into server side workers


pbrant:
Thanks!

PR #129, opened 2017-09-19
cache previous width and position when calculate currentWidth

brucewinger:
The current doBreakText will recalculate the width of the whole phase passed in. The commit caches the calculated previous width and previous position, so it only needs to calculate increased string. It decreases algorithm complexity from O(n^2) to O(n).


pbrant:
Thanks!

pbrant:
Released as 9.1.8

PR #130, opened 2017-10-12
Add page event for writer when creating pdf file

vietnezo:
With ITextRenderer for render pdf files, I notice that there is no PdfPageEventHelper to add custom page evens such as header, footer,... So I 've made this pull request to add page event in ITextRenderer class.
My English is not so good and I 'm sorry for that.
Thank you very much!


adridadou:
is there a reason why this hasn't been merged ? I'm looking at a way to do that

PR #131, opened 2017-11-06
Extract page number from image src URIs

jodersky:
When PDFs are included as images, optionally specify the page number
as a URI fragment in the form "page=<number>".
This follows the convention specified in RFC 3778 [1].

[1] https://tools.ietf.org/html/rfc3778


pbrant:
Thanks!

pbrant:
Released as 9.1.8

jodersky:
Thanks for merging this so quickly!

PR #133, opened 2017-11-08
Removed print stack trace

adamovic-cw:
Hi,

We use this project in our production environment to allow users to define PDF templates, render them and send to their customers. 

Since users can define PDF templates, our log and inbox gets spammed with messages which are caused by printStackTrace call.

I see you have already made a release yesterday, but is there any chance this request gets merged and released soon? I would really appreciate it.


pbrant:
Released as v9.1.9. It should be on Central in a couple of hours.

adamovic-cw:
Thanks a lot. :)

PR #135, opened 2018-01-04
added detection for server relative URLs

Dr4K4n:
I had a problem with css background properties with relative URIs (i.e. `background: url('/image/background.jpg');`).

Flying saucer was trying to load `/image/background.jpg` instead of `http://myserver.com/image/background.jpg` when loading the CSS containing the property from `http://myserver.com/style/print.css`.

I added a method to detect these "server" relative URIs and prefixed the URIs with the host part of the CSS-File URI.


asolntsev (approved):
I am curious why the project doesn't have any tests for this URL magic in CSSParser.java
Sad, sad... :(

But the change seems reasonable for me, so I am going to merge it.

PR #136, opened 2018-01-23
ITextRenderer to allow pageEvents

rik86:
When rendering HTML to PDF using `ITextRenderer` there is no possibility to set the page events (such as header or footers) because the PdfWriter is build inside the **createPDF** method.

Similarly to `_pdfVersion` and `_pdfEncryption` field, there should be a `_pdfPageEvents` field to set events on the PdfWriter before rendering.


rik86:
[ITextRenderer.txt](https://github.com/flyingsaucerproject/flyingsaucer/files/1656328/ITextRenderer.txt)

PR #137, opened 2018-01-23
add _pdfPageEvents to ItextRenderer

rik86:
ITextRenderer to allow pageEvents

When rendering HTML to PDF using `ITextRenderer` there is no possibility to set the page events (such as header or footers) because the PdfWriter is build inside the **createPDF** method.

Similarly to `_pdfVersion` and `_pdfEncryption` field, there should be a `_pdfPageEvents` field to set events on the PdfWriter before rendering.

### DIFF
--- a/flying-saucer-pdf/src/main/java/org/xhtmlrenderer/pdf/ITextRenderer.java
+++ b/flying-saucer-pdf/src/main/java/org/xhtmlrenderer/pdf/ITextRenderer.java
@@ -62,6 +62,7 @@ import org.xhtmlrenderer.util.Configuration;
 import org.xml.sax.InputSource;
 
 import com.lowagie.text.DocumentException;
+import com.lowagie.text.pdf.PdfPageEventHelper;
 import com.lowagie.text.pdf.PdfWriter;
 
 public class ITextRenderer {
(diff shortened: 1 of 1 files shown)


rik86:
Hello @pbrant we would be really keen in having this fix reviewed and released, whenever convenient. Many thanks

pbrant:
Hi @rik86, sorry for the late reply. This should already be possible with `PDFCreationListener`. If you set one with `setListener` on your `ITextRenderer` instance, the `preOpen` method will be called with the current `ITextRenderer`. From there, you can call `getWriter` to make whatever changes you need.

pbrant:
Please do reopen if I'm wrong though.

rik86:
Hi @pbrant the I can call getWriter but that would return a different instance of writer, because `PdfWriter writer = PdfWriter.getInstance(doc, os);`
gets created inside the method `public void createPDF(OutputStream os, boolean finish, int initialPageNo)` before rendering the PDF. I don't know how to reopen the PR

pbrant:
It's being assigned to the `_writer` instance variable on line 328 so it should be available on the `ITextRenderer` passed into the `preOpen` call.

rik86:
Thanks @pbrant for your help I managed to set a PDFCreationListener as you suggested.

PR #138, opened 2018-01-30
docx output

plutext:
This pull request is to see whether there is interest in adding docx output to flying saucer?

By way of background, docx4j has provided XHTML to docx conversion based on a re-packaged flying saucer from late 2011.   XHTML to docx seems to be a fairly common requirement.

(I re-packaged FS back then when FS was dormant, and hadn't gotten around to seeing whether there was appetite for including docx capability in FS-proper.  YellowfinBI then did the work to re-integrate it in FS, which has made this pull request possible).

If there is enthusiasm for XHTML to docx as part of FS, I'd be happy to help clean up this pull request a bit.

To actually convert XHTML to docx, you'd use https://github.com/plutext/docx4j-ImportXHTML/tree/FS2018 which has FS (and docx4j) as a dependency: https://github.com/plutext/docx4j-ImportXHTML/blob/FS2018/pom.xml#L298

docx4j-ImportXHTML's codebase could also be moved here (flying-saucer-docx) if that were necessary/desirable.

Looking forward to discussion/feedback, thanks! .. Jason

### DIFF
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/context/StyleReference.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/context/StyleReference.java
@@ -85,8 +85,29 @@ public class StyleReference {
      * @param userAgent PARAM
      */
     public StyleReference(UserAgentCallback userAgent) {
+        this(userAgent, null);
+    }
+
+    
+    /**
+     * A docx4j extension to standard FS, allowing extra stylesheets
+     * (ie that aren't explicit, either linked or embedded) to be specified 
+     * and applied.  This is useful, for example, for adding CSS which
+     * represent WordML styles in the target package, which can be
+     * used via @class.  OpenDoPE uses this in its XHTML import.
+     * It could also be useful in a web-based editing scenario. 
+     */
+    private StylesheetInfo[] extraCSS;
+    
+    /**
+     * Default constructor for initializing members.
+     *
+     * @param userAgent PARAM
+     */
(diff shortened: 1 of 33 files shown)


asturio:
It's an interesting Frontend feature.

Lozic:
Great!
Would it be easy to adapt to produce xlsx and pptx?

plutext:
pptx: there is already some proof of concept level code at https://github.com/plutext/docx4j-ImportXHTML/tree/master/src/main/java/org/pptx4j/convert/in/xhtml which re-uses DocxRenderer

xlsx: https://github.com/alanhay/html-exporter does this, using FS.  Not sure how complete it is. Its https://github.com/alanhay/html-exporter/tree/master/src/main/java/uk/co/certait/htmlexporter/writer/excel uses POI (but would be straightforward to port to use docx4j/xlsx4j instead; or there might be value in allowing the user to configure it to use either)

pbrant:
Hey @plutext, this looks super cool! FS is still pretty dormant. I merge bug fixes and do releases as people ask me, but there hasn't really been much active development for a very long time. I do try to be quite conservative as to the changes that go into the existing output targets as befitting the age and stage of the project. 

That said, if you'd find it convenient to develop and release this as part of the main repository, that's fine with me. It looks like there are some changes to -core that may cause trouble for other output targets. If you can clean those up (and it's worth your while to do so), I'm happy to merge this, assist with releases, etc.

One other note -- copyright is held by the individual contributor (you or your employer).

GitMensch:
@plutext any update on this?

PR #141, opened 2018-06-15
Refactoring in Logs

Gaboso:
Changed the order of access modifiers
Sorted the imports alphabetically
Added @Override annotation in the required methods
Moved Array designator "[]" to type
Removed unsed imports
Using sigle quotes in indexOf to faster performance
Added type to Lists and Maps


pbrant:
Thanks!

pbrant:

PR #142, opened 2018-06-21
Avoid artifacts due to missing antialiasing

KlausRe:
Setting the clip prior to fill() with the background color results in pixel artifacts for
elements with e.g. border-radius.
Clipping is now only applied when a background image is rendered. 
Without a background image the library will render smooth edges with the Java2DRenderer.


pbrant:
Thanks!

pbrant:

PR #143, opened 2018-07-01
Support box-sizing border-box

balazs-zsoldos:
This is the copy of [the commit](https://github.com/danfickle/openhtmltopdf/pull/143/commits/15dfb2833fed10ba79f63219be62b4dd8ece4291) of the [openhtmltopdf ](https://github.com/danfickle/openhtmltopdf) project. 

I think supporting "box-sizing: border-box" is very important as all CSS libraries start with this setting on every HTML element. Without that, it is almost impossible to set sizes with non-px value (percentage, inch, etc.) and use border or padding anywhere.


pbrant:
Thanks!

pbrant:

PR #144, opened 2018-07-30
Refactoring in TrueTypeUtil

Gaboso:
* reordered imports;
* simplified some method logic;
* added type to map;


pbrant:
Released as v9.1.15

PR #145, opened 2018-07-31
Add version and license badges

Gaboso:
Add version and license badges to README


pbrant:
We're using LGPL 2.1 (which includes the option of using a later version if desired). if you can update the PR to use a 2.1 badge instead, I'd be happy to merge.

Gaboso:
Badge has been updated to use LGPL-2.1 😀

pbrant:
Thanks!

PR #146, opened 2018-07-31
Fixed italic emulation applying to oblique fonts

pixilation:
Like italic fonts, oblique fonts are already slanted, so there's no need to emulate the effect.

This will most commonly be an issue with Helvetica which is loaded internally in ITextFontResolver as the default Sans Serif font.


pbrant:
Thanks. This will be included in the next release.

pbrant:
Released as v9.1.15

PR #150, opened 2018-09-19
Added support for base64 encoded font faces.

atomcat1978:


pbrant:
Thanks! Would you like a release with this change?

atomcat1978:
Hei,

You are welcome:)

Yes, definetely, if possible. Useful in such scenarios when an HTML e-mail attachment and a pdf generation is required the same time. For the mail attachment it is wise to embed the fonts with base64. In such a scenario a common font-face definition file might be used with base64 encoded font faces.

atomcat1978:
Hi,

When can we expect this feature to be released?

PR #153, opened 2018-12-03
Add dispose for each create.

OndrejSpanel:
Each Graphics2D.create should be paired with a dispose call. Missing this causes serious issues with `GLGraphics2D` (see https://github.com/brandonborkholder/glg2d/)


pbrant:
Thanks. This will be included in the next release later today.

pbrant:

PR #154, opened 2018-12-06
Rounded corners without borders added twice into the path because of bad starting angle.

OndrejSpanel:
The Flying Saucer was generating duplicate segments for rounded corners when border width was zero. Each side of the rectangle generated full 90 degrees rounded corner for both left and right side. This has caused significant performance issues in awt Area constructor sometimes. I have a simple layout where just creating single awt.Area from a shape took over 1500 ms! With multiple rounded elements per page this causes delays of multiple seconds.

The time was spent inside of function `sun.awt.geom.AreaOp.pruneEdges`  called  from `sun.awt.geom.AreaOp.calculate`. The function is probably unable to handle the path which contains duplicate segments properly.

I have changed the code so that only 45 degrees are generated in this case. I have checked the samples to verify it did not cause any new issues in the border or area rendering.

### DIFF
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/render/BorderPainter.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/render/BorderPainter.java
@@ -107,18 +107,20 @@ public class BorderPainter {
             sideWidth = bounds.width-(1+scaledOffset)*(widthScale)*(border.left()+border.right());
         }
         Path2D path = new Path2D.Float();
-        
-        float angle = 90;
+
+        float fullAngle = 90;
+        float defaultAngle = fullAngle / 2;
+        float angle = defaultAngle;
         float widthSum = props.getTop() + props.getLeft();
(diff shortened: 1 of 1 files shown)


pbrant:
Hi Ondrej, thanks for this. Do you have anything else coming? If not, I'm happy to do a release with this and the other MR you opened recently.

OndrejSpanel:
I  do not know. I am working on using Flying Saucer to render the UI layer in an OpenGL based game (via [GLG2D ](https://brandonborkholder.github.io/glg2d/)layer), and solving issues as I encounter them. I think I already have it working now, but it is possible something else will come up. There is no need to synchronize release with me, however, as I can deploy my own version into a private nexus repository anytime I need. I am making PRs here so that the issues are fixed here in long-term (and as a way to at least partially show the gratitude and pay back - I have found the Flying Saucer very useful for my purpose, thanks a lot - it is a tremendous effort).

pbrant:

PR #155, opened 2018-12-13
Allow border collapse even on paginated tables.

lapo-luchini:
Inspired from https://github.com/danfickle/openhtmltopdf/issues/97, this simply removes an exception from the code and allows border collapse even on paginated tables.
It works and is much better looking / less of an unexpected change in output, at least to me.
(I wonder why that exception was in place in the first place, though)


pbrant:
Merging for release later today.

It's been ages since I've worked on this, but my recollection is that I just wasn't sure it would work well. Boarding collapsing works by picking a winning border for each side. I think I was worried that some cells would end up with missing border edges. In any case, if it appears to work, I'm happy to merge.

pbrant:

PR #156, opened 2019-03-05
Fixed unnecessary line break in tables

philipparndt:
With flyingsaucer v9.1.14 the table rendering is broken for some use-cases.
I use this project to render markdown documents within a eclipse application so this is a essential use-case.

Example html document:
[table.zip](https://github.com/flyingsaucerproject/flyingsaucer/files/2930165/table.zip)

Rendering result in v9.1.13:
after

Rendering result in v9.1.14 (and following):
before

Rendering result in Chrome (for reference):
chrome

The issue was introduced with the following change:
https://github.com/flyingsaucerproject/flyingsaucer/commit/e8a5b35fd9b56481a68d38e101728ce1f1e30269

The idea of the patch is to restore the original behavior when not using "border box".


pbrant:
Thanks much for the fix. There are a couple of other patches I'd to like to include, but I'll get a release out with this change by tomorrow.

philipparndt:
Thank you, that was fast! I'll integrate the new version in my project as soon as it is released.

pbrant:

PR #158, opened 2019-03-15
nth-child condition fix

datb-com:
Great to see that flyingsaucer is still being maintained, and now with OpenPDF - excellent!

I found a bug in support for the nth-child CSS condition.  This suggested correction aligns to the formal representation and matches Firefox, Safari and Chrome.
See screenshots in the attached for the pre and post flyingsaucer behaviour and the same CSS in browsers.

2019-03-15_07-13-22 flying saucer nth child test

### DIFF
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/css/newmatch/Condition.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/css/newmatch/Condition.java
@@ -409,15 +409,53 @@ abstract class Condition {
             // getPositionOfElement() starts at 0, CSS spec starts at 1
             int position = treeRes.getPositionOfElement(e)+1;
 
-            position -= b;
 
-            if (a == 0) {
-                return position == 0;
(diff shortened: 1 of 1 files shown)


pbrant:
Thanks! Do you need a release with this change? I'm happy to do one if so, but if not, I'll bundle it into the next one (which, depending on the number of patches submitted, is probably a few months away).

datb-com:
Peter

No, I don't need a release urgently, I'll use my locally modified copy.

We've been using flyingsaucer for quite a few years and hadn't been keeping up to date.  It was great yesterday to be able to update FS and move from iText to OpenPDF with very little issue. 

I'm about to look into a pagination issue.  I sometimes get a blank page when I do a page-break-before:always.  My theory is that if the page that follows is a full page it does another page break.  If I have any luck with this I'll raise another PR.

Thanks for a great library.

Jason


Jason Dore
Technical director
datb
07970 663669
01572 720767

PR #159, opened 2019-04-15
fixes issue with outline generation

peterm0x:
Outline is not correctly generated when writing multiple documents into single PDF. Bookmarks of the first document are (augmented and) reused for subsequent documents. PDF writer ultimately fails with 
```
RuntimeException: The page 6 was requested but the document has only 5 pages.
```
due to references to missing pages.


pbrant:
Thanks! Would you like a release with this change? (if not, I'll roll it into the next patch release [whenever that is])

andreasrosdal:
A new release of the flyingsaucer-openpdf build would be perfect!

Snesi:
I'm not sure if this will help. But we've been running into the `The page 6 was requested but the document has only 5 pages` issue when converting an html page to pdf that contains `page-break-*` css properties.

PR #160, opened 2019-05-10
Fix behavior of :first pseudo selector on named pages

Gondy:
Currently, `:first` pseudo selector is working ONLY on very first PDF page. But according to https://www.w3.org/TR/css-page-3/#syntax-page-selector :first pseudo selector should be applied also on first page of named pages (sections):

```
@page { ... }
@page :left { ... }
@page :right { ... }
@page LandscapeTable { ... }
@page CompanyLetterHead:first { ... } /*  identifier and pseudo page. */
```
Problem is only with `:first` selector. `:right` and `:left` are working fine.

Problem is described also here:
https://stackoverflow.com/questions/36737431/flying-saucer-pdf-using-first-with-a-named-page 

Solution is keep track of current page number of actual named page. For example, if we are on named page "section1" then start counting pages of this named pages block. If pending page number == 1 and there is rule `@page section1:first {...}`  it should apply. No matter what page number of whole PDF it is.

### DIFF
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/layout/Layer.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/layout/Layer.java
@@ -81,6 +81,8 @@ public class Layer {
     
     private Set _pageSequences;
     private List _sortedPageSequences;
+
+    private Map _pendingPagesMap = new HashMap();
     
     private Map _runningBlocks;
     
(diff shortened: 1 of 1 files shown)


pbrant:
My reading of the spec is that :first only applies to the very first page.

https://www.w3.org/TR/css-page-3/#first-pseudo

Is that modified somewhere else?

PR #163, opened 2019-09-13
Added maxlength and multiline to text form field

KalhariL:
Following two changes are added. Both were verified and are working

At the moment maxlength property is ignored and the resulting text areas have no limit to the text input. Added max length limit to text form field. This is a todo in this file.  

If a height is defined for text input, make it a multiline input. At the moment even if a area gets painted to match the height it's always a single line input.


pbrant:
This was released as part of v9.1.19, which should be on Maven Central in the next couple of hours. Thanks for the PR.

PR #164, opened 2019-10-28
Add support for relative links to filesystem resources

netheosgithub:
Flying Saucer is not generating relative links in the form of:

``<a href="documents/mydoc.txt">doc</a>``

(We are generating a PDF linking to relative non-embedded assets)

I found the same issue raised here:
https://groups.google.com/forum/#!searchin/flying-saucer-users/$20links%7Csort:date/flying-saucer-users/zbGoaROnY34/bKtpNCi1K5QJ
but the suggested solution is not supported by the current implementation.

Checking the source, I saw a 
```
if (uri.length() > 1 && uri.charAt(0) == '#') {
...
} else if (uri.indexOf("://") != -1 || uri.startsWith("mailto:")) {
...
}
```

I'd like to suggest fixing the ``else if`` to just a `else` to support any uri input form including relative paths.


netheosgithub:
If the fix acceptable, it would be nice to have a release very soon.

netheosgithub:
@pbrant considering your merge/release policy, would it be possible for you to have a look?  thanks for your time

PR #167, opened 2019-11-14
Replaced javax.xml.bind.DatatypeConverter with java.util.Base64

maltalex:
The jaxb APIs which include `javax.xml.bind.DatatypeConverter` are considered "Java EE" and are no longer contained by default in Java 9+. Since `DatatypeConverter` was only used for Base64 decoding, I've replaced it with `java.util.Base64` which should always exist.

### DIFF
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/util/FontUtil.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/util/FontUtil.java
@@ -1,8 +1,8 @@
 package org.xhtmlrenderer.util;
 
-import javax.xml.bind.DatatypeConverter;
 import java.io.ByteArrayInputStream;
 import java.io.InputStream;
+import java.util.Base64;
 import java.util.logging.Level;
 
 public class FontUtil {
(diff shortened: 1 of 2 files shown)


maltalex:
The jaxb APIs which include `javax.xml.bind.DatatypeConverter` are considered "Java EE" and are no longer contained by default in Java 9+. Since `DatatypeConverter` was only used for Base64 decoding, I've replaced it with `java.util.Base64` which should always exist.

pbrant:
Thanks. If this is causing you issues, I'd be happy to do a release with this change. Otherwise it will be rolled into the next patch release after a few other PRs have accumulated (probably in a few months).

maltalex:
Hi @pbrant 
It's not causing any major problems, just had to explicitly include the missing classes. There's no rush to release a new version.

astiob:
I’d like to request a release if the offer still stands.

This causes `NoClassDefFoundError` at run-time on JRE 11 with no warnings at compile-time, and to work around it one presumably needs to include an unobvious runtime dependency (either `org.glassfish.jaxb:jaxb-runtime` or `org.eclipse.persistence:org.eclipse.persistence.moxy`) to their own project that uses Flying Saucer, as well as spend some time to figure out what the error means in the first place.

Of course, it should be noted that `java.util.Base64` only exists since Java 8. I’m not sure what JRE compatibility Flying Saucer aims for, but if it includes both Java 7 and Java 9 (which is the first version that _hides_ JAXB by default, before it’s removed completely in Java 11), then extra code (or another third-party decoder) will be needed to make it work across all versions.

pbrant:
This was released as
https://github.com/flyingsaucerproject/flyingsaucer/releases/tag/v9.1.20

It's on Bintray now. It should be on Maven Central in a couple of hours.

PR #170, opened 2020-06-27
Guard against race condition

DaveJarvis:
Assign `styleMap` to local method variable (`localMap`) to avoid possibility of `styleMap` becoming `null` during method execution, otherwise resulting in an NPE.

### DIFF
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/layout/SharedContext.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/layout/SharedContext.java
@@ -553,7 +553,7 @@ public class SharedContext {
     }
 
     public CalculatedStyle getStyle(Element e, boolean restyle) {
-	Map localMap = styleMap;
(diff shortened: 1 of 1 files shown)


DaveJarvis:
@pbrant -- could we get this one merged?

DaveJarvis:
@pbrant -- bump?

pbrant:
I think this will only be relevant if using SharedContext concurrently across multiple threads. Does that sound right given given your use of it?

This isn't how the code was intended to be used and it wouldn't surprise me if there are other issues with doing so. It can be used repeatedly (after calling .reset()) across a single thread though.

That said, the patch as written shouldn't cause any issues either so I'm fine with it.

DaveJarvis:
I don't recall the exact circumstances around when I encountered the NPE, sorry. Merging would be great, because then I could get replace my custom build with the one from this repo.

pbrant:
OK, sounds good.

pbrant:
Released 9.2.1

PR #171, opened 2020-06-27
Maven compiler plugin version

DaveJarvis:
Increase Maven Compiler Plugin version to prevent build failures on recent versions of Java and Maven.


pbrant:
Released 9.2.0

PR #173, opened 2020-08-23
Make VerifyGlyphExists compile using IntelliJ

DaveJarvis:
Changes:

* Add reference to `org.w3c.dom` in pom.
* **BorderPropertySet.java** -- Organize imports (had compile issues without org.w3c.dom).
* **VerifyGlyphExists.java** -- Compare line 100 (old) and line 110 (PR); the codepoint was always 0, which looks like a bug. Other updates modernize the syntax and apply IDE-recommended changes, including try-with-resources. This closes the opened font's file stream, which is technically another bug but one that would never affect the application in practice.
* **ITextFontResolver.java** -- Organize imports.

### DIFF
--- a/flying-saucer-core/pom.xml
+++ b/flying-saucer-core/pom.xml
@@ -37,6 +37,11 @@
       <version>4.10</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.w3c</groupId>
+      <artifactId>dom</artifactId>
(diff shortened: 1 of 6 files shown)


asolntsev (approved):

asolntsev:
@DaveJarvis Thanks. Will be released as FlyingSaucer 9.2.3

PR #175, opened 2021-03-16
Replicating the commit 3e47546

ottobeli:
Replicating the commit 3e47546 from rafortis to this project. It is updated on flying-saucer-pdf but it wasn´t replicated into flying-saucer-pdf-itext5.
This change avoid an IllegalArgumentException thrown by PdfName in iTextpdf project when the element has more than 127 characters.


pbrant:
Thanks. Would you like a release with this change?

ottobeli:
> 
> 
> Thanks. Would you like a release with this change?

Yes please.

pbrant:
Released as v9.1.21

PR #178, opened 2021-05-25
Support for no page margin mode when html is converted to PDF

zhaokun-12:
Support for no page margin mode when html is converted to PDF


pbrant:
Closing out old PRs. @zhaokun-12, if you'd still like this in, please reopen. It looks it might be incomplete though (?).

PR #179, opened 2021-10-28
remove absolute path prefix for licenses

nickbreen:
when copied into source jar target by maven resources plugin


pbrant:
Released 9.2.0

PR #180, opened 2022-04-24
upgrade bouncycastle to 1.70

urld:


urld:
CVE-2020-15522 in bcprov 1.64 looks bad on depency-check report when using flying-saucer-pdf-osgi

urld:
Thx for merging this. However, i think bcprov 1.70 has been outdated again for a while now. You may  want to check the latest version. I think that would be 1.76.

pbrant:
Thanks. I've updated to 1.76.

pbrant:
Released 9.2.0

PR #181, opened 2022-11-22
Default Accept header value is invalid according to RFC 7231

Pitupitupit:
Default Accept header value is invalid according to RFC 7231. 
https://bugs.openjdk.org/browse/JDK-8163921
Fixed in java 19 which sets its value to `*/*`

Requests got blocked after firewall upgrade with rule that checks Accept header.


pbrant:
Released 9.2.0

PR #182, opened 2022-12-21
Support for radio buttons

mnegrean:
Uncommented the code for radio buttons. The background color is static but at least the radio buttons appear on the generated pdf. 

Before:

Screenshot from 2022-12-21 10-30-41

After:

Screenshot from 2022-12-21 10-30-51


mnegrean:
@pbrant not sure if you are still merging PRs or releasing new versions to maven central, but it would help me tremendously to have support for radio buttons. Thanks in advance!

pbrant:
Released 9.2.0

PR #184, opened 2023-02-10
Make CJKFont optional

sc-mickael:
ITextFontResolver init is very slow on my computer and server (8000ms). After some research, it seems that createInitialFontMap() method take too much time. The cause is the cjkfonts which I did not use.
Is it possible to make this loading optional ?
Thanks a lot

### DIFF
--- a/flying-saucer-pdf/src/main/java/org/xhtmlrenderer/pdf/ITextFontResolver.java
+++ b/flying-saucer-pdf/src/main/java/org/xhtmlrenderer/pdf/ITextFontResolver.java
@@ -42,11 +42,17 @@ import java.util.stream.Stream;
 public class ITextFontResolver implements FontResolver {
     private Map _fontFamilies = createInitialFontMap();
     private Map _fontCache = new HashMap();
-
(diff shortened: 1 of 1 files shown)


asolntsev (approved):

asolntsev:
@sc-mickael Upvote for this change. Loading Asian fonts takes ~5 seconds - which is too slow when we run unit-tests or re-generate PDF while developing templates.

The initial intention was to load these fonts ONLY if file `iTextAsian.jar` is available on the class path. So it was NOT loaded by default for most users. 

But nowadays file `cjkfonts.properties` is contained in file `openpdf-1.3.11.jar`, and this cannot be removed from classpath.

pbrant:
Released 9.2.0

sc-mickael:
Thanks a lot ! I tried it and it seems to work!

asolntsev:
For the record: later I converted parameter `withCjkFonts` to subclass `CJKFontResolver` in https://github.com/flyingsaucerproject/flyingsaucer/pull/204

PR #200, opened 2023-09-17
[refactoring] use generics

asolntsev:


flying-saucer-core/src/main/java/org/xhtmlrenderer/layout/SharedContext.java
```diff
@@ -318,16 +316,16 @@ public NamespaceHandler getNamespaceHandler() {
 
     public void addBoxId(String id, Box box) {
         if (idMap == null) {
-            idMap = new HashMap();
+            idMap = new HashMap<>();
```
DaveJarvis, line 319: Would it be more thread-safe to initialize `idMap` upon construction? Then there's no race-condition between the check for `null` and the `put()` call. If the map needs to be set to `null`, it can be `clear()`ed instead?

This would apply to lines 325-328 as well.

flying-saucer-core/src/main/java/org/xhtmlrenderer/layout/SharedContext.java
```diff
@@ -65,7 +65,7 @@ public class SharedContext {
 
     private boolean interactive = true;
 
-    private Map idMap;
+    private Map<String, Box> idMap;
```
DaveJarvis, line 68: Consider instantiating the `Map` to avoid potential race-condition.

flying-saucer-pdf/src/main/java/org/xhtmlrenderer/pdf/ITextOutputDevice.java
```diff
@@ -925,18 +925,15 @@ private void writeBookmarks(RenderingContext c, Box root, PdfOutline parent, Lis
     }
 
     private void writeNamedDestinations(RenderingContext c) {
-        Map idMap = getSharedContext().getIdMap();
+        Map<String, Box> idMap = getSharedContext().getIdMap();
         if ((idMap != null) && (!idMap.isEmpty())) {
```
DaveJarvis, line 931: If `isMap` is guaranteed to never be null, would that simplify this entire method?

DaveJarvis (changes requested):

flying-saucer-core/src/main/java/org/xhtmlrenderer/layout/SharedContext.java
asolntsev: Yes, it's a good point. 
Most of these changes, I did just by applying IDEA shortcuts etc. It's quite hard to review & improve all of them with a single PR. My philosophy is "at least do not make it worser". :)

flying-saucer-core/src/main/java/org/xhtmlrenderer/css/parser/property/AbstractPropertyBuilder.java
```diff
@@ -99,7 +101,7 @@ protected void checkIdentOrLengthType(CSSName cssName, CSSPrimitiveValue value)
 
     protected void checkIdentOrNumberType(CSSName cssName, CSSPrimitiveValue value) {
         int type = value.getPrimitiveType();
-        if (type != CSSPrimitiveValue.CSS_IDENT && type != CSSPrimitiveValue.CSS_NUMBER) {
+        if (type != CSSPrimitiveValue.CSS_IDENT && type != CSS_NUMBER) {
```
DaveJarvis, line 104: The IDE didn't use a static import for both constants?

asolntsev: Small steps, my friend, small steps. :)
IDE suggests million improvements. I don't want to apply all of them at once. One-by-one.

flying-saucer-core/src/main/java/org/xhtmlrenderer/layout/SharedContext.java
asolntsev: done

flying-saucer-core/src/main/java/org/xhtmlrenderer/layout/SharedContext.java
asolntsev: done

flying-saucer-pdf/src/main/java/org/xhtmlrenderer/pdf/ITextOutputDevice.java
asolntsev: done

PR #205, opened 2023-09-26
Remove repackaged JAXB DOM dependency

pbrant:


asolntsev (approved):

asolntsev:
Hm... Initially I thought this JAXB dependency was needed to compile the project on newer JDK (11 or 17?).
But now I see that the project compiles nicely without JAXB on all JDK versions 8..21

PR #220, opened 2023-11-02
Update to OpenPDF 1.3.32. Set OpenPDF profile as default.

andreasrosdal:
Update to OpenPDF 1.3.32. 
Set OpenPDF profile as default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ New releases of Flying Saucer are distributed through Maven. The available artif
 
 * `org.xhtmlrenderer:flying-saucer-core` - Core library and Java2D rendering
 * `org.xhtmlrenderer:flying-saucer-pdf` - PDF output using iText 2.x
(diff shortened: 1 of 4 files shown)


asolntsev:
@andreasrosdal It's not that simple. Currently, the release process of FlyingSaucer is a bit complicated (I need to switch to a separate branch `openpdf` to build that specific artifact `flyingsaucer-openpdf`). I need to rework & simplify the release script first. Hope to do it soon.

asolntsev:
After thinking a bit, I came to an idea: what if we totally remove support for itext, and just replace "itext" implementation with "openpdf"? Then we don't need to refactor the release scripts.

At least I migrated one production project from itext to openpdf, and it has been working in production for few months already. No problems encountered.

asolntsev:
@andreasrosdal Finally, I did this change in https://github.com/flyingsaucerproject/flyingsaucer/pull/245.
Will be released as FlyingSaucer 9.4.0 this year. :)

PR #277, opened 2024-02-28
simplify API for creating PDF

asolntsev:
now user needs to write less code to generate PDF from HTML

This commits adds 2 new apis:
1. class Html2Pdf   (a single method for generating PDF from HTML)
2. renderer.createPDF(doc, os)   (instead of old sequence `setDocument`, `layout`, `createPDF`)

The work continues...

### DIFF
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/resource/XMLResource.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/resource/XMLResource.java
@@ -48,8 +48,10 @@ import javax.xml.transform.sax.SAXSource;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.Reader;
+import java.io.StringReader;
 import java.lang.ref.Reference;
 import java.lang.ref.SoftReference;
+import java.net.URL;
 import java.util.Queue;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.logging.Level;
@@ -77,6 +79,10 @@ public class XMLResource extends AbstractResource {
         super(source);
     }
 
+    public static XMLResource load(URL source) {
+        return load(new InputSource(source.toString()));
+    }
+
     public static XMLResource load(InputStream stream) {
         return XML_RESOURCE_BUILDER.createXMLResource(new XMLResource(stream));
     }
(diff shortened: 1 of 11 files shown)


andreasrosdal:
Is is possible to add a XHTML Validator Regex Fixer Class as part of this, something like this:


https://chat.openai.com/share/c7f9c46e-1874-491a-8fc9-04d78518a8fe

```
import java.util.regex.*;

public class XHTMLValidator {
    public static void main(String[] args) {
        String inputHTML = "<html><head><title>Sample</title></head><body><p>Paragraph</p><div><span>Span Text</span></div></body></html>";
        
        // Regular expression to match HTML tags
        String regex = "<\\s*([\\w]+)(.*?)>(.*?)<\\s*/\\s*\\1\\s*>";

        // Create a Pattern object
        Pattern pattern = Pattern.compile(regex);

        // Create a Matcher object
        Matcher matcher = pattern.matcher(inputHTML);

        // Use a StringBuffer to store the fixed XHTML
        StringBuffer fixedXHTML = new StringBuffer();

        // Loop through matches and fix invalid XHTML
        while (matcher.find()) {
            String startTag = matcher.group(1);
            String endTag = "</" + startTag + ">";
            String content = matcher.group(3);
            
            // Append valid start tag, content, and end tag
            matcher.appendReplacement(fixedXHTML, "<" + startTag + matcher.group(2) + ">" + content + endTag);
        }

        // Append the remainder of the inputHTML
        matcher.appendTail(fixedXHTML);

        // Print the fixed XHTML
        System.out.println(fixedXHTML.toString());
    }
}
```

asolntsev:
@andreasrosdal Yes, generally we could add some validation/html cleanup.
But using regular expressions doesn't seem to be a good idea for this purpose: https://medium.com/thecyberfibre/stop-parsing-x-html-with-regular-expression-2cf13215b411

andreasrosdal:
How about using Jsoup to fix invalid XHTML?
https://jsoup.org/

In general I agree that regular expressions should not be used to fix bad HTML. However, we could use something, and at least regular expressions is one of the possible alternatives. Jsoup is probably better.

andreasrosdal:
https://github.com/jhy/jsoup/

asolntsev:
@andreasrosdal Yes, JSoup sounds good. 
Could you pelase share some example of such "invalid XHTML" needing a cleanup? I am trying to understand what problem we want to solve.

andreasrosdal:
Here are some examples of html changes had to be made to make the html XHTML valid for flying saucer:
image
image
image
image

asolntsev:
@andreasrosdal Thank you for the samples. Some of these could be really replaced automatically (e.g. `<link>` -> `<link></link>`), but some others are actually invalid, and FS should throw exception in these cases (e.g. `<<i class="">`).

andreasrosdal:
I wish that FS would handle these cases, and not throw exception, in the same way that Firefox and Chrome is fault tolerant for invalid HTML in many cases.

PR #280, opened 2024-03-06
Adding F flag to make links pdf/a-1a conform

haramon:
Flying Saucer is not generating links from a tags that are PDF/A-1A conform:

`<a href="https://www.google.at/">My link</a>`

The veraPDF validation`of [Rule 6.5.3-2](https://github.com/veraPDF/veraPDF-validation-profiles/wiki/PDFA-Part-1-rules#rule-653-2) fails with following error:

`An annotation dictionary shall contain the F key. The F key's Print flag bit shall be set to 1 and its Hidden, Invisible and NoView flag bits shall be set to 0.`

This fix adds the `F` key to the annotation of generated links, so the pdf passes the PDF/A-1A conformity check.

---

If the fix is acceptable, it would be nice to have a release soon. If version 9.6.0 cannot be released yet, it is also ok cherry-picking this commit to 9.5.2 and release from there.

Thanks for your time :)

### DIFF
--- a/flying-saucer-pdf-itext5/src/main/java/org/xhtmlrenderer/pdf/ITextOutputDevice.java
+++ b/flying-saucer-pdf-itext5/src/main/java/org/xhtmlrenderer/pdf/ITextOutputDevice.java
@@ -31,6 +31,7 @@ import com.itextpdf.text.pdf.PdfContentByte;
 import com.itextpdf.text.pdf.PdfDestination;
 import com.itextpdf.text.pdf.PdfImportedPage;
 import com.itextpdf.text.pdf.PdfName;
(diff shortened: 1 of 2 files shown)


asolntsev (approved):
I have no idea what it means, but seems reasonable. :)

asolntsev:
@haramon Released FlyingSaucer 9.6.0. Thank you for the contribution! 

P.S. Probably we should add VeraPDF to test dependencies, and validate all generated PDFs?

haramon:
Thanks, for the fast review and release 👍

Yeah, it would be convenient to cover the veraPDF validation with tests.

PR #302, opened 2024-03-20
Fix center image and glyph list item marker

haramon:
The image and glyph list item marker are not perfectly vertically centered.

Before: The center of the text **baseline** and the **ascent** is used as the baseline (which does not consider the line height and therefore places the marker a little bit more to the top).

Fix: The center of the **ascent** and the **descent** is used as the baseline.

also see [wiki](https://en.wikipedia.org/wiki/Baseline_(typography))

---
**Results:**
The results are screenshots of the [demo sample list.xhtml](https://github.com/flyingsaucerproject/flyingsaucer/blob/main/flying-saucer-examples/src/main/resources/demos/list.xhtml).

Before:
list_before

Fix:
list_after

### DIFF
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/render/ListItemPainter.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/render/ListItemPainter.java
@@ -59,15 +59,13 @@ public class ListItemPainter {
         MarkerData.ImageMarker marker = markerData.getImageMarker();
         FSImage img = marker.getImage();
         if (img != null) {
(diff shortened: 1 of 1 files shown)


haramon:
hi @asolntsev
it would be great if you can take a look at this PR

thanks in advance :)

pbrant (approved):

asolntsev (approved):

haramon:
thanks a lot 👍

asolntsev:
@haramon Thank you for the contribution. I've released the fix as FS 9.7.2

PR #318, opened 2024-04-30
Add a test for the embedding of fonts declared in the css

perNyfelt:
Add a test that verifies that the `-fs-pdf-font-embed: embed;` indeed results in an embedded font in the PDF.


asolntsev:
Thank you @perNyfelt for the test!
I think we could also add this example to [documentation](https://flyingsaucerproject.github.io/flyingsaucer/r8/guide/users-guide-R8.html)

NB! When I run your test, it works even without `-fs-pdf-font-embed: embed;` line in css.

PR #324, opened 2024-05-26
Remove lib directory

andreasrosdal:
Remove lib directory, since this is not in use, outdated, and not needed.


asolntsev (approved):
Great!

Issue #282, opened 2024-03-08
Support for HTML 5


andreasrosdal:
FS should support HTML 5.

https://www.w3.org/TR/2011/WD-html5-20110405/
https://html.spec.whatwg.org/

Possibly some implementation details can be copied from:
https://github.com/openhtmltopdf/openhtmltopdf/

rbri:
Maybe https://github.com/HtmlUnit/htmlunit-neko is of help here.
This
* parses html into a (valid) dom tree
* has no other dependencies
* is currently in use by HtmlUnit and therefore in many projects
* active developed

Because my time is limited i can't provide a impl but i will support this if you like...

andreasrosdal:
This is magnificent @rbri - would you please be inspired by the PR in #327 and possibly submit a new pull request for us adding html parsing using both https://github.com/HtmlUnit/htmlunit-neko and https://github.com/jhy/jsoup - supporting both in Flying Saucer with real native support for both parsers and letting the users choose?

andreasrosdal:
@rbri I would like to encourage you to make a pull request which allows using the neko-htmlunit html parser in Flying Saucer, in a default way without any hassle configuration, because this html parser is clearly better than the current xml sax parser. This could make it much easier to recommend using Flying Saucer to the developers in the company I work, because at the moment FS is no good because it only supports strict xhtml and developers look for alternatives to FS now.

rbri:
@andreasrosdal PR is there ;-) if guess we need some discussion about the right way to do it (maybe a service and a different subproject?)

Issue #279, opened 2024-03-04
FlyingSaucer can cleanup invalid html


asolntsev:
FlyingSaucer could be more tolerant to minor errors in html source. 
Instead of throwing an exception, say, for non-closed html5 tags, it could close them automatically.

For example, we could use JSoup for job. 

Triggered in discussion https://github.com/flyingsaucerproject/flyingsaucer/pull/277

DarkTyger:
I agree with @pbrant : If an application wants to clean up HTML before passing it to FlyingSaucer, that's the application's job. A module would be good, but I already have a specific version of JSoup included with my application for various reasons (including, but not limited to, clean up before passing to Flying Saucer). Having potentially two versions of JSoup would unnecessarily bloat my application.

Can we close this issue?

asolntsev:
@DarkTyger Yes, we could close the issue and leave this problem for FS users to solve by themselves.
But it's a common request, and many users want to solve this problem.
So it's better to solve it once in FS than make people re-invent the wheel every time.

> An optional plugin (or module) would be good.

This is exactly what I was think about: a separate plugin / module / something like that. 

> Having potentially two versions of JSoup would unnecessarily bloat my application.
You will never have two versions of JSoup. Both Maven and Gradle always leave only one version in classpath.

PR #327, opened 2024-05-26
Use Jsoup to parse HTML

andreasrosdal:
Use JSOUP to parse HTML. 
https://jsoup.org/

Deprecate XMLResource. 
Add HTMLResource.

Jsoup is a HTML parser, so it could work better to parse HTML than the SAX parser currently in use.
I consider this a proposal, which I think is a step in the right direction. However, I am not fully sure that I understand all the consequences of this yet. So I would like to propose this change, maybe it will be accepted. Related: #279 and #282. 

Further, Jsoup is a HTML parser, and most users of Flying saucer will be expecting HTML syntax to be valid, not just XHTML. The current parser in Flysing Saucer is a SAX based XML parser which will throw exceptions if there input is not valid XHTML.

And importantly, we need to make sure that this doesn't introduce any XSS or HTML based vulnerabilities.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,10 @@
 
 Flying Saucer is a pure-Java library for rendering arbitrary well-formed XML 
 (or XHTML) using CSS 2.1 for layout and formatting, output to Swing panels, 
-PDF, and images.
+PDF, and images. 
+
+The new [Jsoup](https://jsoup.org/) based HTML parser will allow supporting HTML 5 syntax
+and features in Flying Saucer, with a goal of implementing full HTML5 support over time. 
 
 Comprehensive documentation available in [The Flying Saucer User's Guide](https://flyingsaucerproject.github.io/flyingsaucer/r8/guide/users-guide-R8.html).
 
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/resource/HTMLResource.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/resource/HTMLResource.java
@@ -1,0 +1,104 @@
+/*
+ * {{{ header & license
+ * Flying Saucer - Copyright (c) 2024
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public License
+ * as published by the Free Software Foundation; either version 2.1
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.	See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+ * }}}
+ */
+package org.xhtmlrenderer.resource;
+
+import org.jsoup.Jsoup;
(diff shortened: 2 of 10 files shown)


andreasrosdal:
@jhy Does this use of Jsoup look fine?

flying-saucer-core/src/main/java/org/xhtmlrenderer/resource/HTMLResource.java
```diff
@@ -0,0 +1,93 @@
+/*
+ * {{{ header & license
+ * Flying Saucer - Copyright (c) 2024
```
pbrant, line 3: Copyrights should be owned by individual contributors.

flying-saucer-core/src/main/java/org/xhtmlrenderer/resource/HTMLResource.java
```diff
@@ -0,0 +1,93 @@
...
+        }
+    }
+
+    public static HTMLResource load(InputStream stream) {
+        return new HTMLResource(stream);
+    }
+
+    public static HTMLResource load(Reader reader) {
+        try {
+            InputStream stream = convertReaderToInputStream(reader);
```
pbrant, line 68: Can jsoup read from a Reader directly? It would be nice to avoid the conversion.

README.md
```diff
@@ -5,7 +5,10 @@
 
 Flying Saucer is a pure-Java library for rendering arbitrary well-formed XML 
 (or XHTML) using CSS 2.1 for layout and formatting, output to Swing panels, 
-PDF, and images.
+PDF, and images. 
+
+The new [Jsoup](https://jsoup.org/) based HTML parser will allow supporting HTML 5 syntax
+and features in Flying Saucer, with a goal of implementing full HTML5 support over time.
```
pbrant, line 11: The enthusiasm is awesome, but I worry this is so far away as to be borderline misleading. 

Supporting HTML5 syntax as input (which is already possible without code changes) is one thing. It doesn't make it any easier to implement the vast array of new features that have been added to HTML/CSS. For example, CSS Grid and Flexbox are huge, complicated features in and off themselves. They are simply out of reach as things stand now.

pbrant (changes requested):
Please address comments and move jsoup support to a separate module.

andreasrosdal:
Yes, I can address these comments and move the jsoup support to a separate module.

How about a separate module for htmlunit-neko also, in a similar way. As pointed out by @rbri in https://github.com/flyingsaucerproject/flyingsaucer/issues/282#issuecomment-2133685493 it seems that htmlunit-neko is also a quite capable html parser. So I am thinking it could be useful to support both jsoup and htmlunit-neko parsers, but I'm not sure how at this time.

flying-saucer-core/src/main/java/org/xhtmlrenderer/resource/HTMLResource.java
andreasrosdal: The main Jsoup class has no Reader method here: https://jsoup.org/apidocs/org/jsoup/Jsoup.html
However, there is https://jsoup.org/apidocs/org/jsoup/parser/Parser.html#parseInput(java.io.Reader,java.lang.String)

andreasrosdal:
Will submit an update.

rbri:
> How about a separate module for htmlunit-neko also, in a similar way. As pointed out by @rbri in https://github.com/flyingsaucerproject/flyingsaucer/issues/282#issuecomment-2133685493 it seems that htmlunit-neko is also a quite capable html parser. So I am thinking it could be useful to support both jsoup and htmlunit-neko parsers, but I'm not sure how at this time.

Great idea - will try to support this

PR #330, opened 2024-05-28
Update Configuration programmatically

balrok:
This feature would be helpful for us, so we can make sure that in tests and runtime the config is always the same on all machines.


asolntsev:
@balrok Do I correctly understand that the new approach allows users to change the `Configuration` on the fly at any moment?
I did the same mistake years ago in Selenide, and I know: users **will abuse it**. 
They will call `Configuration.setProperty` before, after and between tests. They will call `Configuration.setProperty` with different values in parallel tests etc. 

Maybe it's a better approach to allow users create a configuration file, e.g. `flyingsaucer.conf`?
Then you can commit such a file in `src/main/resources`, and the same configuration will be used in tests and production.

PR #333, opened 2024-06-02
simple impl to try to use https://github.com/HtmlUnit/htmlunit-neko SAX parser (see #282)

rbri:
There seems to be no dynamic way to add another try, did this simple hack.

Hope someone with more knowledge about this lib comes up with some better ideas....

### DIFF
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/resource/XMLResource.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/resource/XMLResource.java
@@ -66,10 +66,12 @@ public class XMLResource extends AbstractResource {
     private Document document;
     private static final XMLResourceBuilder XML_RESOURCE_BUILDER;
     private static boolean useConfiguredParser;
+    private static boolean useHtmlUnitCyberNekoParser;
 
     static {
         XML_RESOURCE_BUILDER = new XMLResourceBuilder();
         useConfiguredParser = true;
+        useHtmlUnitCyberNekoParser = true;
     }
 
     private XMLResource(InputStream stream) {
@@ -115,8 +117,8 @@ public class XMLResource extends AbstractResource {
 
     public static XMLReader newXMLReader() {
         XMLReader xmlReader = null;
-        String xmlReaderClass = Configuration.valueFor("xr.load.xml-reader");
 
+        String xmlReaderClass = Configuration.valueFor("xr.load.xml-reader");
         //TODO: if it doesn't find the parser, note that in a static boolean--otherwise
         // you get exceptions on every load
         try {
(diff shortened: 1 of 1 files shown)


andreasrosdal:
Thank you, my 2 cents as a user of Flying Saucer, would be to add htmlunit-neko as a dependency in FS pom.xml and make it the default html parser, because it is simply much better than the current xml sax parser.

andreasrosdal:
@asolntsev Hi, what's your opinion on this?

pbrant:
I'm afraid it's pretty much a non-starter. There is a very high probability that it would break a large number of users without warning.

An HTML parser parsing an XML document won't always create the same DOM as an XML parser parsing an XML document.

Making it easy to swap out the XMLReader used globally sounds like a good idea though.

rbri:
@pbrant my guess is, the s sourcer

andreasrosdal:
https://github.com/openhtmltopdf/openhtmltopdf
https://github.com/danfickle/openhtmltopdf

Note these two forks, which have taken steps in supporting html by default. I think FS should move in the same direction.

pbrant:
@rbri That's not quite accurate. I'd describe Flying Saucer as a W3C DOM renderer that, by default, parses input as XML (not XHTML).

For an example of how the parsing rules differ consider this HTML5/XHTML document which is also valid XML:

```html
<html>
  <body>
    <p>
      one
      <div>two</div>
      three
    </p>
  </body>
</html>
```

An HTML5/XHTML parser will produce the DOM equivalent of the following (taken from DevTools):

```html
<html><head></head><body>
    <p>
      one
      </p><div>two</div>
      three
    <p></p>
  

</body></html>
```

These two DOMs won't render the same in Flying Saucer even with the default stylesheet and since their internal structure differs, user stylesheets might also match differently.

pbrant:
> 
> Note these two forks, which have taken steps in supporting html by default. I think FS should move in the same direction.

A fork starting with zero users has a lot more flexibility than a project with hundreds of thousands of downloads a month.

andreasrosdal:
My focus here is to encourage and discuss ways of supporting the parsing of HTML instead of, or in addition to, xhtml. This is motivated by the problems I have encountered with using FS to parse xhtml. Possibly we can be inspired by the FS forks, and see how they implemented the html parser there. And if the html parser works better and will make FS work better overall, in addition to the existing xhtml parser possibly, then it could be worth to create a new major version which adds support for a html parser. So I hope to encourage solutions to move towards supporting html parsing. Otherwise the FS forks will take over usage quickly for new projects and for active software projects which needs to create PDF files from HTML.

andreasrosdal:
Adding an easy way to use htmlunit-neko as sax parser would be an improvement to FS. 
Eventually over time htmlunit-neko could be made the default html parser.

 How can we make this task progress towards inclusion?

andreasrosdal:
https://github.com/openpdfsaucer/openpdfsaucer

Here's the fork with zero users, using htmlunit-neko as html parser.

PR #357, opened 2024-07-26
Overcoming Table Cell and Line Splitting Challenges

syluna:
While using Flying Saucer, we encountered two issues: the first one concerns the rendering of cells with overlapping rowspan and colspan which draws a border in the middle. The second issue is that during PDF generation, the table headers are correctly repeated, but the table rows are incorrectly split.

Attached, you will find an illustration of the XHTML and the two issues fixed by this code are highlighted in red.
Before fix issues :
Sans titre

After fix issues :
xhtmlrender
cut row and border


pbrant:
Please resubmit without the whitespace changes. It looks like you might have your editor set to use tabs instead of spaces. It should use four space indents.

PR #358, opened 2024-07-27
Overcoming Table Cell and Line Splitting Challenges

syluna:
While using Flying Saucer, we encountered two issues: the first one concerns the rendering of cells with overlapping rowspan and colspan which draws a border in the middle. The second issue is that during PDF generation, the table headers are correctly repeated, but the table rows are incorrectly split.

Attached, you will find an illustration of the XHTML and the two issues fixed by this code are highlighted in red.
Before fix issues :
Sans titre

After fix issues :
xhtmlrender
cut row and border

### DIFF
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/newtable/TableCellBox.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/newtable/TableCellBox.java
@@ -41,859 +41,859 @@ import java.util.List;
 import java.util.Set;
 
 public class TableCellBox extends BlockBox {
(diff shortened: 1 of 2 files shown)


asolntsev (approved):
@pbrant The code looks good to me (though it's quite hard to understand what exactly it does).

@syluna Couldn't you write some test? I understand that such visual things can be tested only manually, but still...

P.S. Please always commit formatting changes (even small ones like `top.setWidth((top.width()+1)/2);`) in a separate PR. Otherwise, it's quite hard to understand what was actually changed in the behavour.

PR #359, opened 2024-08-05
Update pattern of headings

pedroo002:
Modify pattern so that headings won't have an upper limitation for their level


asolntsev (approved):

pedroo002:
In our solution we are making pdf documents in which heading levels are not limited and we use html as intermediate documents.

PR #378, opened 2024-09-03
Add automatic module name to MANIFEST.MF in flying-saucer-core

OpenHelios:
This change will make it easier to use flying-sourcer-core in a module based project.

### DIFF
--- a/flying-saucer-core/pom.xml
+++ b/flying-saucer-core/pom.xml
@@ -62,6 +62,7 @@
         <configuration>
           <instructions>
             <Bundle-SymbolicName>org.xhtmlrenderer.flying.saucer.core</Bundle-SymbolicName>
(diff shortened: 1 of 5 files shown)


asolntsev (approved):
I have no idea what effect this change takes :)

But why not. Seems reasonable. :)

asolntsev:
@OpenHelios Why you added `<Automatic-Module-Name>` only to one module "flying-saucer-core"?
FS has several other modules. Most often, people want to use module "flying-saucer-pdf", not core.

OpenHelios:
I have amended my commit to add the automatic module name to the other public Maven modules. In the case of flying.sourcer.pdf I used the maven-jar-plugin to add the entry into the MANIFEST.MF file.

asolntsev (approved):

PR #381, opened 2024-09-05
an attempt to fix 3-parameter @page size parameter

jelinj8:
I wanted to create a PDF for a label printer with landscape orientation AND a custom size which evidently isn't supported.
```
@page {
size: 40mm 20mm landscape;
}
```
This is an attempt to fix that (+ fixed one typo in original code reporting error on incorrect parameter value).

I'm unable to test this properly right now as I'm unable to compile the library with Java 17 (still using FS 9.4.1 with Java 1.8), but it compiled OK with slight modifications for the old version, this is to get it fixed when I'll be able to upgrade and to give something back, thanks.


jelinj8:
And I've made a similar error as I've fixed (error message on line 163 - taking value1 instead of value3).

asolntsev:
@jelinj8 Wait, you don't need to re-create PR if you made a typo.
Just do "amend commit" and "force push".

PR #383, opened 2024-09-06
Page size 3value fix

jelinj8:
I wanted to create a PDF for a label printer with landscape orientation AND a custom size which evidently isn't supported.

```
@page {
   size: 40mm 20mm landscape;
}
```

This is an attempt to fix that (+ fixed one typo in original code reporting error on incorrect parameter value).

I'm unable to test this properly right now with current version as I'm unable to compile the library with Java 17 (still using FS 9.4.1 with Java 1.8), but it compiled OK and works (with slight modifications for the old version), this is to get it fixed when I'll be able to upgrade and to give something back, thanks.
Succesfully created portrait label on landscape paper 40x20mm...

### DIFF
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/css/parser/property/SizePropertyBuilder.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/css/parser/property/SizePropertyBuilder.java
@@ -148,34 +148,34 @@ public class SizePropertyBuilder extends AbstractPropertyBuilder {
                 throw new CSSParseException("Invalid value for size property", -1);
             }
         }  else if (values.size() == 3) {
-			PropertyValue value1 = (PropertyValue) values.get(0);
-			PropertyValue value2 = (PropertyValue) values.get(1);
-			PropertyValue value3 = (PropertyValue) values.get(2);
+            PropertyValue value1 = (PropertyValue) values.get(0);
+            PropertyValue value2 = (PropertyValue) values.get(1);
+            PropertyValue value3 = (PropertyValue) values.get(2);
 
-			checkInheritAllowed(value3, false);
+            checkInheritAllowed(value3, false);
 			
-			if (isLength(value1) && isLength(value2) && value2.getPrimitiveType() == CSSPrimitiveValue.CSS_IDENT) {
-				if (value1.getFloatValue() < 0.0f) {
-					throw new CSSParseException("A page dimension may not be negative", -1);
-				}
+            if (isLength(value1) && isLength(value2) && value2.getPrimitiveType() == CSSPrimitiveValue.CSS_IDENT) {
+                if (value1.getFloatValue() < 0.0f) {
+                    throw new CSSParseException("A page dimension may not be negative", -1);
+                }
 
-				if (value2.getFloatValue() < 0.0f) {
-					throw new CSSParseException("A page dimension may not be negative", -1);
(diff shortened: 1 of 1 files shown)


jelinj8:
Tested on latest version - works
[label.pdf](https://github.com/user-attachments/files/16906850/label.pdf)
configured by
```
@page {
	margin: 0;
	size: 40mm 20mm landscape;
}
```

flying-saucer-core/src/main/java/org/xhtmlrenderer/css/parser/property/SizePropertyBuilder.java
```diff
@@ -94,7 +94,7 @@ public List<PropertyDeclaration> buildDeclarations(
             } else {
                 throw new CSSParseException("Value for " + cssName + " must be a length or identifier", -1);
             }
-        } else { /* values.size == 2 */
+        } else if (values.size() == 2) { /* values.size == 2 */
```
asolntsev, line 97: Is this comment is still needed?
It just duplicates the code.

flying-saucer-core/src/main/java/org/xhtmlrenderer/css/parser/property/SizePropertyBuilder.java
```diff
@@ -147,6 +147,35 @@ public List<PropertyDeclaration> buildDeclarations(
             } else {
                 throw new CSSParseException("Invalid value for size property", -1);
             }
+        }  else if (values.size() == 3) {
+            PropertyValue value1 = (PropertyValue) values.get(0);
```
asolntsev, line 151: Can't we have at least some tests for this code?

It should be easy:

```java
class SizePropertyBuilderTest {
    private final SizePropertyBuilder builder = new SizePropertyBuilder();

    @Test
    void buildDeclarations() {
        PropertyValue cssValue = new PropertyValue(CSS_NUMBER, 2.0f, "foo");
        List<PropertyDeclaration> result = builder.buildDeclarations(COUNTER_INCREMENT, List.of(cssValue), 42, false);
        assertThat(result).hasSize(1);
        assertThat(result).containsExactly(new PropertyDeclaration(COUNTER_INCREMENT, cssValue, false, 42));
    }
}
```

asolntsev (changes requested):

flying-saucer-core/src/main/java/org/xhtmlrenderer/css/parser/property/SizePropertyBuilder.java
jelinj8: cleaned, sorry (modification made in plaintext editor)

flying-saucer-core/src/main/java/org/xhtmlrenderer/css/parser/property/SizePropertyBuilder.java
jelinj8: I'll try to prepare that, I'm a bit struggling with building the lib in Eclipse...

asolntsev:
@jelinj8 Thanks for the PR. 
I've released [FS 9.9.3](https://github.com/flyingsaucerproject/flyingsaucer/releases/tag/v9.9.3) and [added few tests](https://github.com/flyingsaucerproject/flyingsaucer/commit/671d4383ac0d22e3e3a785284e754db3256abc3a).

PR #391, opened 2024-09-21
JSoup HTML parser in separate module

andreasrosdal:
JSoup HTML parser in separate module:

* A new flying-saucer-core-nextgen module with JSoup HTML parser has been added.  https://jsoup.org/ 
This new core module is for adding new features and supporting modern HTML standards. flying-saucer-core is for stable legacy xhtml parsing. flying-saucer-core and flying-saucer-core-nextgen are interchangeable.

* Flying-saucer-swt: Replace bundled jar files with proper dependencies from maven central.
* XMLResource has been renamed to HTMLResource.

This is my second attempt at improving the HTML parsing support in Flying Saucer. With these changes the Flying Sacuer browser is able to parse the HTML of e.g. Google.com, in a better way than the existing Xhtml based Sax parser. I humbly realize that improving the HTML parser support in Flying Saucer is a massive complex task. This has been an interesting learning experience, and I will not be surprised if this PR is rejected. Still I think it might be a step in the right direction: separating the stable flying-saucer-core from an experimental flying-saucer-core-nextgen module which uses JSoup. If accepted, then I would like to continue improving CSS parsing support and adding JavaScript support.

image


pbrant:
Hey Andreas, thanks for the PR. I appreciate the effort that went into it. I'm afraid it's kind of an example of "hunting mice with an elephant gun" though.

It would be less invasive to add a service interface to allow a user to swap out the DOM parser implementation used by the Swing-based mini-browser (either auto-configured by the presence of the module or explicitly swapped out through configuration). I think I may have suggested this before.

Supporting additional CSS properties is an almost entirely orthogonal problem to the DOM parser in use. This could be done while using an XML, an XHTML, or HTML5 parser to create the DOM.

We do have some experience with copy-n-pasted modules. The old flying-saucer-pdf-itext5 module was effectively a clone of flying-saucer-pdf with package changes and minor API updates. 

To put it bluntly, it was a disaster. It had already bitrotted rather badly by the time it was deleted as most contributed fixes only touched flying-saucer-pdf. I'm quite happy that Andrei had the courage to delete it.

It's awesome that you'd like to start experimenting with supporting more CSS properties and adding JavaScript. It is a hugely ambitious task. I'd suggest starting that effort in a separate fork to see how it goes.

asolntsev:
@andreasrosdal @pbrant In fact, we already have an example showing how to use JSoup to parse HTML:

https://github.com/flyingsaucerproject/flyingsaucer/blob/main/flying-saucer-examples/src/test/java/org/xhtmlrenderer/pdf/PdfFromInvalidHtmlTest.java

But yes, we could improve it even more by service loader mechanism...

Issue #409, opened 2024-10-10
Remove all "main" methods


asolntsev:
I would like to remove all "main" methods from FlyingSaucer project.
It seems to be legacy not used by anyone.

Some of them could be converted to automated tests, while others just don't work anymore.

For example:
<img width="1335" alt="image" src="https://github.com/user-attachments/assets/4d253600-df68-444a-8a5c-04ba8f79bb5a">

@pbrant

pbrant:
Sounds good to me. BrowserStartup may be the only one with any utility
anymore.

asolntsev:
@pbrant Which `BrowserStartup` do you mean? I see two of them:

<img width="507" alt="image" src="https://github.com/user-attachments/assets/a03c5966-9002-465d-87b3-fa0f71c31ab9">

And none is working:
```
Exception in thread "main" java.lang.UnsatisfiedLinkError: Cannot load 32-bit SWT libraries on 64-bit JVM
	at org.eclipse.swt.internal.Library.loadLibrary(Unknown Source)
        ...
	at org.eclipse.swt.widgets.Display.<init>(Unknown Source)
	at org.xhtmlrenderer.demo.browser.swt.BrowserStartup.main(BrowserStartup.java:27)
```

pbrant:
I think it was this
one ./flying-saucer-examples/src/main/java/org/xhtmlrenderer/demo/browser/BrowserStartup.java.

asolntsev:
@pbrant Sorry, but I don't understand how to use `BrowserStartup`. What problems does it solve? Can't we replace it by an automated tests?

pholdings:
> @pbrant Sorry, but I don't understand how to use `BrowserStartup`. What problems does it solve? Can't we replace it by an automated tests?

I just verified that this still works. Here's one approach that worked for me

- from project root, mvn clean install
- from flyingsaucer-examples, mvn compile
- to get a working classpath, mvn dependency:build-classpath
- build the call using that classpath: java -cp $THAT_CLASSPATH:target/classes org.xhtmlrenderer.demo.browser.BrowserStartup

This opens up a small Swing/desktop application that shows Flying Saucer as if it were a really simple web browser. In the "demo" menu you can walk through some ancient demo pages showing how we could cover different parts of CSS.

You can also browse XHTML websites, like http://redwriteblue.com/tags/tutorialxhtml.html (just an example I found in a search right now). 

[...]

PR #419, opened 2024-10-12
#409 refactor XRLogger

asolntsev:
* make field non-nullable: `XRLog.loggerImpl`
* remove unneeded lazy evaluation (and field `XRLog.initPending`)
* remove unneeded method `XRLog.main`

Ideally, I would like to replace all this custom logging code by (de-facto standard) SLF4J.


pbrant (commented):
Looks good to me. I agree it would be nice to use slf4j too.

Issue #411, opened 2024-10-10
Regression in Unordered List layout when rendering via OpenPDF v2.x


TarqusMiltonium:
### Description

I have a set of HTML documents that get converted from HTML to PDF via Flyingsaucer+OpenPDF. Upon upgrading from OpenPDF v1.3.26 and flying-saucer-pdf-openpdf v9.1.22 to OpenPDF v2.0.3 and flying-saucer-pdf-openpdf v9.9.2, unordered lists are no longer rendered correctly when the `<li>` element contains any `margin-top|bottom` or `padding-top|bottom` styling -- The first bullet still lines up with the first line of text, but subsequent bullets are misaligned relative to their respective lines.

I have prepared a minimal test repo to demonstrate this.

### Expected Behaviour
The bullets should be aligned with their text.
This works in OpenPDF v1.3.26.

### Current Behaviour
The first bullet lines up with the first line of text, but subsequent bullets are misaligned relative to their lines.
See screenshots for details.

### Screenshots
| Version | Screenshot |
| ---- | ---- |
| OpenPDF v1.3 | ![image](https://github.com/user-attachments/assets/602a6d6a-c9b2-454b-a516-522d0847e961) |
| OpenPDF v2.0 | ![image](https://github.com/user-attachments/assets/986e557b-1420-4ae9-973b-39bea5514aa3) |

Resultant PDF Files also attached.
[output_openPDF_v1.3.pdf](https://github.com/user-attachments/files/17323011/output_openPDF_v1.3.pdf)
[output_openPDF_v2.0.pdf](https://github.com/user-attachments/files/17323012/output_openPDF_v2.0.pdf)

### Steps to Reproduce
See repo at https://github.com/TarqusMiltonium/openPdfRenderingUpgradeExample

Simply run commits [98b458e](https://github.com/TarqusMiltonium/openPdfRenderingUpgradeExample/commit/98b458edbcd8bf76c82e19d3c86b19cd3efa378b) and [f9e4f1e](https://github.com/TarqusMiltonium/openPdfRenderingUpgradeExample/commit/f9e4f1e252147c282a0b6e49800fd77825275e1d) respectively.

### Workaround
I did find a workaround: adding the following CSS restores correct layout, however this must be added to every single html document, so it's not a great answer.
```css
        ul {
            list-style-type: none;
        }

        li::before {
            content: '• ';
            margin-left: -1em;
            padding-right: 0.7em;
            font-size: 1.3em;

        }
[...]

PR #420, opened 2024-10-13
#411 fix center li items with margin or padding

haramon:
fyi: I could not start `org.xhtmlrenderer.demo.browser.BrowserStartup` on the current main branch (had to checkout older commit to test the fix). Not quite sure if it is only on my machine 🤔 - had no time to take a closer look.


asolntsev (approved):
I have no idea what's happening here, but it helped :)

PR #442, opened 2024-11-12
Fix PDFAsImage scaling and uses of FSImage.scale for immutability changes

JamesScrase-PortSwigger:
https://github.com/flyingsaucerproject/flyingsaucer/commit/2ce6f9f05a058485185af39d7757b7ded8033280#diff-d210bc82c50b9ada2d6d90ecc8bb3688eaa4df211c1dd891615590558b732759R65 seems to have broken scaling of `` tags which reference `.pdf` files - `ITextReplacedElementFactory` wasn't updated to use the new return value.

This PR updates the remaining usages of `FSImage.scale` to use the returned `FSImage` instead of the existing object, and passes through the unscaled width and height to the new `PDFAsImage` in `PDFAsImage.scale`.


asolntsev (approved):
Wow! Thank you. My bad.
I had to annotate `scale` method with `@CheckReturnValue` and double-check all usages.

PR #492, opened 2025-04-10
Fix border rounding issues and overlap singularities

OndrejSpanel:
This PR fixes an issue in Java2D backend, which manifests with some rounded corner configurations. The issue is the corner may be broken by a white line, because the shapes do not connect perfectly.

Rounded corners are also used in the demo file splash.xhmtl instead of using a bitmap with rounded corners.

Note: the code is quite old. I do not remember exact details, I am not in process of syncing my fork with this repository and I therefor submit fixes here I have there.

### DIFF
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/render/BorderPainter.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/render/BorderPainter.java
@@ -47,10 +47,10 @@ public class BorderPainter {
      * @return A Path that is all sides of the round rectangle
      */
     public static Path2D generateBorderBounds(Rectangle bounds, BorderPropertySet border, boolean inside) {
-        Path2D path = generateBorderShape(bounds, TOP, border, false, inside ? 1 : 0, 1);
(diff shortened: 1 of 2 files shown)


asolntsev:
@OndrejSpanel Thank you for the fix. I can merge it, but can you explain what's the problem you are trying to solve? Maybe some example?

I tried your splash.html with old and new code, and the result is the same.

OndrejSpanel:
I experienced some round corner rendering issues a few years ago. I do not remeber what exact border configuration was necessary to show them. I can try to find it if you think it is important.

asolntsev:
> I experienced some round corner rendering issues a few years ago. I do not remember what exact border configuration was necessary to show them. I can try to find it if you think it is important.

I don't know, maybe not. Since I don't see any differences with my eyes, maybe I can just trust and and merge this PR. :)

OndrejSpanel:
I have added a few examples into the borders page.

Before the fix:

image

After the fix:

image

asolntsev (approved):
It's hard for me to check for the moment, but I trust the change. :)

PR #493, opened 2025-04-10
Revieve linear gradients and opacity PR

OndrejSpanel:
Attempt to revive #22

It works fine in my fork, it "just" needs resolving possible conflicts.


OndrejSpanel:
Demo screenshot:

image

asolntsev:
@OndrejSpanel Thank you for the contribution. 
Implementing CSS3 features is a valuable direction for FS. 
What's your plan? Going to resolve those conflicts?

OndrejSpanel:
I think it should be ready now. If it makes review or merging easier, I could prepare a squashed version.

OndrejSpanel:
Note: my effort concentrates on `Graphics2D` rendering. While opacity seems to be implemented for PDF, I did not test it. The `drawLinearGradient` function in `ITextOutputDevice` is just a stub. I do not use the PDF functionality and I am not familiar with necessary APIs. If anyone want to see this, feel free to take over.

asolntsev:
@OndrejSpanel Well... I personally don't care about Graphics2D, I use FlyingSaucer to generate PDFs. 
So I can just trust your PR and merge it. :)

Usually I prefer to _rebase_ PR, but in this case this option is not available because of conflicts:
image


So what do you prefer, should we _squash_ all 70 commits into one, or is it better to create a merge commit?
image

OndrejSpanel:
> I use FlyingSaucer to generate PDFs.

Regarding PDF support: while gradients are not implemented in the PDF backend, they should not be difficult to add if anyone is interested. And the opacity should already work, only I did not test it for PDFs.

asolntsev:
@OndrejSpanel Thank you. After looking at the PR, I decided to merge the original PR - let's keep its history. In case of any troubles, it will be easier to investigate and understand why some changes were done.

ledfordtj:
@asolntsev did the PDF support for linear gradient get added before this was merged? This package works amazing for my use case, except I need support for linear gradient support for PDFs as well. I am happy to help out where I can, but I think I will need some direction.

asolntsev:
No, I think PDF support was not implemented. 
You are welcome. :)

OndrejSpanel:
I think the change is not related to gradients and should not be sneaked-in together with them.

Other than that, I think nowadays it makes sense to have antialiasing always on.

I think anyone can provide PR for that.

PR #523, opened 2025-06-22
SVG support

asolntsev:
Now it's possible to use SVG images in HTML templates:

```html
<img src="classpath:/star.svg" />

<img src="/img/star.svg" />

<div class="container" style="background: center url('classpath:/star.svg') no-repeat; background-size: 50%;" />

<div class="container" style="background: center url('/img/star.svg') no-repeat; background-size: 40%;" />
```


oclay1st:
Hi @asolntsev , 
Thanks for this, but can you make this feature optional?. This introduce new dependencies and should not be tied to pdfs.

asolntsev:
@oclay1st What problem are you solving? Yes, now comes one more dependency - so what? Why is this a problem?

oclay1st:
I mean something like what  [OpenHtmlToPDF](https://github.com/danfickle/openhtmltopdf/tree/open-dev-v1/openhtmltopdf-svg-support/src/main/java/com/openhtmltopdf/svgsupport) did should be more appropriate

asolntsev:
@oclay1st Sorry, I still don't understand what problem you are solving? Extra dependency? I checked - it's 4MB. Is this the problem? Or something is actually broken?

Do you mean to create a separate module like `flying-saucer-pdf-svg` to avoid this 4MB dependency?

oclay1st:
In my opinion, there's no reason to have Batik as a direct dependency and this svg support is tied to pdfs. Furthermore, it increases the size and difficulty of compiling this library as a native image with GraalVM. But hey, this is just my suggestion. You can keep it as it is. Good luck

louis-jaris:
Thank you for this work @asolntsev 🙏 

I did not understood why SVGs weren't working when I was making my docs using flyingsaucer 🙏 

Back then, I had to convert my SVG into PNGs, which are greatly impacting the PDF size -- I'm curious to see if now my (tens of thousands, and soon hundred of thousands) PDF will be positively impacted by this change 🙏

asolntsev:
Yes, we also had to convert SVG images to PNG and even encode them to base64 data urls. :)

But I am not sure your PDFs size will decrease because during PDF generation, FlyingSaucer still converts svg to png. :)

Issue #530, opened 2025-07-02
page-break-inside: avoid not working


xlaussel:
Hi,

page-break-inside: avoid is not working sometime.

I have an html code like this for producing a pdf file:

```html
<body>
    <div>
     ...... Lot of stuff
    </div>
    <div style="page-break-inside: avoid; border: solid 1px red; height: 800px;page-break-before: always;"></div>
    <div style="page-break-inside: avoid; border: solid 1px red; height: 300px;">Breaked!! But shouldn't</div>
</body>
```
The last div is breaked across 2 pages.

When I try without the first div like this it is OK:

```html
<body>
    <div style="page-break-inside: avoid; border: solid 1px red; height: 800px;page-break-before: always;"></div>
    <div style="page-break-inside: avoid; border: solid 1px red; height: 300px;">Not breaked</div>
</body>
```

I use the last flying-saucer-pdf version: 9.13.0

I don't understand why the page is breaked in the first example. 
It really depends on the stuff in the first div, sometime it is OK sometimes not depending on it.
The "`page-break-before: always`" in the second div should have reset all the paginination stuff so the code before should not impact it.
 I suspect that the `<ul> and <li>` tags before cause the problem but I am not sure.
Here is the full buggy case (sory but I did'nt manage to reproduce the bug with a simpler code):

```html



<body>

    <div class="document">

    <div class="condition">

        <div class="article ">
            <p class="title">Propriété du matériel</p>
            <p>Le MATERIEL restant la propriété exclusive du LOUEUR, le LOCATAIRE ne pourra en disposer, en partie ou intégralement, de quelque manière que ce soit, sans l’accord préalable et écrit du LOUEUR.<br/>
                En cas d’atteinte directe ou indirecte par quiconque audit droit de propriété, il doit en aviser immédiatement le LOUEUR, prendre à ses frais toutes mesures pour faire connaître ledit droit et faire cesser ladite atteinte, notamment par obtention d’une mainlevée.<br/>
                Le LOCATAIRE devra informer tout Mandataire judiciaire de l’existence du contrat de location et lui communiquer l’identité et l’adresse du LOUEUR, notamment en cas de procédure collective.<br/>
                A la demande du LOUEUR, le LOCATAIRE doit apposer et garder, sur le MATERIEL, la plaque de propriété du LOUEUR. Les frais de publicité du contrat et de fabrication de la plaque de propriété sont à la charge du LOCATAIRE.<br/>
[...]

PR #536, opened 2025-07-11
issue #530 page-break-inside: avoid not working fix

xlaussel:
Correction of the #530 issue.
There was a problem in the `BlockBoxing::layoutContent` function.
I did it coherent with the `BlockBoxing::relayoutRun` function :
`mayCheckKeepTogether` is now  reverted to its initial value (was always  set to false before when only a page clear was needed )


asolntsev (approved):

asolntsev:
@xlaussel FlyingSaucer 9.13.1 has been released.

PR #547, opened 2025-08-10
enable anti-aliasing by default

asolntsev:
this restores change that was done in commit e5d6c83ccb and later occasionally reverted in 63f62a625dad.

See https://github.com/flyingsaucerproject/flyingsaucer/pull/493#issuecomment-3168506429


OndrejSpanel:
That commit also changed the default value:

`Configuration.valueAsFloat("xr.text.aa-fontsize-threshhold"` ...

Maye change this as well here?

OndrejSpanel (approved):

PR #563, opened 2025-09-09
Disable leftover debug System.out write

Xyaren:
This log statement is visible when rendering.


asolntsev (changes requested):
@Xyaren Thanks! But why not just remove this line?

PR #586, opened 2025-10-21
ITextRenderer: Enable PDF compression and specify PDF creator and producer

andreasrosdalw:
Improvements to ITextRenderer:
* Enable compression by default. This will result in smaller PDF files.
* Specify PDF producer and creator. This will allow inspecting which software was used to produce a PDF file.

### DIFF
--- a/flying-saucer-pdf/src/main/java/org/xhtmlrenderer/pdf/ITextRenderer.java
+++ b/flying-saucer-pdf/src/main/java/org/xhtmlrenderer/pdf/ITextRenderer.java
@@ -81,6 +81,7 @@ public class ITextRenderer {
     private String pdfProducer = "Flying Saucer 10 with OpenPDF 3.";
     private String pdfCreator = "Flying Saucer 10 with OpenPDF 3.";
     private int compression = 9;
+    private boolean compressionEnabled = true;
 
     private final SharedContext _sharedContext;
     private final ITextOutputDevice _outputDevice;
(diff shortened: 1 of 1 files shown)


andreasrosdal:
A new release of Flying Saucer with this would be nice, I think compression of PDF files can help reduce the file size of generated PDF files.

asolntsev:
I recommend adding "Implementation-Version" attribute to OpenPdf's MANIFEST file.
Like this: https://github.com/selenide/selenide/blob/a64826af0ff123b61990d11243e66e1f12abd823/gradle/publish-module.gradle#L15

Then users can always get its current version:
https://github.com/selenide/selenide/blob/a64826af0ff123b61990d11243e66e1f12abd823/src/main/java/com/codeborne/selenide/webdriver/WebDriverFactory.java#L122

andreasrosdal:
In OpenPDF, the version is written to **org/openpdf/text/version.properties** by the Maven build process.

The version is used here: https://github.com/LibrePDF/OpenPDF/blob/master/openpdf-core/src/main/java/org/openpdf/text/Document.java#L105

And it looks like this:
> bundleVersion=3.0.0

So perhaps Flying Saucer could read the OpenPDF version.properties and show that version in the PDF file producer string in ITextRenderer, in addition Flying Saucer could write it's version as Implementation-Version to it's own MANIFEST file, and include that in the PDF file producer string in ITextRenderer also.

asolntsev:
@andreasrosdal FS 10.0.3 has been released.

PR #623, opened 2025-12-18
Enhance HTML5 element support in Flying Saucer core and expand regression tests

andreasrosdalw:
Enhance HTML5 element support in Flying Saucer core and expand regression tests

See: https://github.com/flyingsaucerproject/flyingsaucer/issues/589

This PR adds default CSS support for HTML5 elements to the Flying Saucer core, ensuring standard tags render correctly without manual configuration.
Changes
Block Elements: Set section, article, header, footer, nav, aside, main, figure, details, and summary to display: block.
Inline-Block Elements: Set <canvas>, <video>, <progress>, and <meter> to display: inline-block.
Semantic Styling: Added default background for <mark>.
Regression Tests: Added Html5SectioningElementsTest.java to verify default display values and ensure author CSS can still override them.
Developed with GitHub Copilot and verified against web standards to ensure predictable layout for modern XHTML documents.

Change done with GitHub copilot using GTP5. Note that I don't fully understand the consequences, but I think this approach of using AI to improve Flying Saucer might be a good idea.

### DIFF
--- a/flying-saucer-core/src/test/java/org/xhtmlrenderer/simple/extend/Html5SectioningElementsTest.java
+++ b/flying-saucer-core/src/test/java/org/xhtmlrenderer/simple/extend/Html5SectioningElementsTest.java
@@ -93,7 +93,7 @@ class Html5SectioningElementsTest {
         CalculatedStyle style = styleFor(tagName);
         assertEquals(IdentValue.INLINE_BLOCK, style.getDisplay(),
                 "Expected <" + tagName + "> to be inline-block by default");
-    }
+    } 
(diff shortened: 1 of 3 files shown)


asolntsev:
@andreasrosdal Looks reasonable, but how we can be sure it works well?
We would need to some sample HTML templates using all these tags, right? And check (at least visually) that the resulting PDF looks good.

andreasrosdal:
Yes, a HTML5 testcase document for HTML5 verification and visual inspection could be possible. Something for 2026.

PR #661, opened 2026-04-27
  Add flying-saucer-html5-core module with HTML5 parsing and rendering support

andreasrosdalw:
- Introduces two new Maven modules: flying-saucer-html5-core (HTML5-aware Java2D renderer) and flying-saucer-html5-pdf (HTML5-aware PDF renderer).
  Both are now listed in the parent reactor and documented in the README.
  - flying-saucer-html5-core uses htmlunit-neko https://github.com/HtmlUnit/htmlunit-neko (org.htmlunit:neko-htmlunit:4.21.0) for HTML5 parsing instead of jsoup. The Html5Parser class wraps neko's DOMParser to produce a W3C Document directly (no intermediate DOM conversion), with element/attribute names normalised to lowercase for Flying Saucer's CSS selector matching.
  - Java2DRenderer gains a protected createNamespaceHandler() factory method so that Html5Java2DRenderer can inject Html5NamespaceHandler (with HTML5
  semantic element support) without duplicating render logic.
  - flying-saucer-examples is updated to depend on flying-saucer-html5-pdf instead of the separate core + PDF pair; jsoup version is now managed centrally in the parent POM.
  - Tests from flying-saucer-core that belong to the shared layout/CSS subsystem are relocated into flying-saucer-html5-core.
  
  (This was created using AI, and is an idea / starting-point for discussing how to modernize Flying Saucer)


asolntsev:
I also planned to use AI to implement html5/css3 features. 

Great!
But why do we need a separate module? Isn't it simpler just to add html5 support to the existing modules?
I personally need only "flying-saucer-pdf" module.

PR #665, opened 2026-05-18
ContentTypeDetectingInputStreamWrapper: tolerate UTF-8 BOM in isSvg()

stanley-leong:
## Summary

`ContentTypeDetectingInputStreamWrapper.isSvg()` does a byte-exact comparison of the first 4 bytes against `"<svg"` / `"<?xm"`. A BOM-prefixed SVG (`EF BB BF` + `<svg…`) fails detection, so the library falls through to raster decoding, returns a null `FSImage`, and the image slot either renders blank or surfaces as a load failure depending on the user agent.

This patch consumes a leading UTF-8 BOM inside the wrapper's constructor so `firstBytes` captures the real first glyph and the magic-byte comparison works as expected. UTF-16 BOMs aren't handled (rare in practice for SVG / XML written by hand).

## Repro

```java
byte[] bom  = { (byte) 0xEF, (byte) 0xBB, (byte) 0xBF };
byte[] body = "<svg><metadata/></svg>".getBytes(UTF_8);
byte[] bytes = concat(bom, body);
var stream = detectContentType(new ByteArrayInputStream(bytes));
assertThat(stream.isSvg()).isTrue();   // fails on main; passes with this patch
```

SVGs saved by Notepad on Windows, some Vim/Emacs configs, or auto-generated by tooling commonly include a leading UTF-8 BOM. Hitting this is purely a function of who produced the SVG asset, not of how the application loads it.

## Test plan

- [x] New `isSvg_withUtf8Bom` and `isSvg_withUtf8Bom_andXmlHeader` tests in `ContentTypeDetectingInputStreamWrapperTest`
- [x] Existing tests in the same file pass unchanged

### DIFF
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/util/ContentTypeDetectingInputStreamWrapper.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/util/ContentTypeDetectingInputStreamWrapper.java
@@ -34,9 +34,29 @@ public class ContentTypeDetectingInputStreamWrapper extends BufferedInputStream 
 
     private final byte[] firstBytes;
 
+    private static final byte[] UTF_8_BOM = { (byte) 0xEF, (byte) 0xBB, (byte) 0xBF };
+
     private ContentTypeDetectingInputStreamWrapper(InputStream source) throws IOException {
         super(source);
+        skipUtf8BomIfPresent(this);
         this.firstBytes = readFirstBytes(this, MAX_MAGIC_BYTES);
+    }
+
+    /**
+     * If the wrapped stream begins with a UTF-8 BOM (EF BB BF), consume it
+     * so {@link #firstBytes} captures the real first glyph. Otherwise the
+     * magic-byte comparison fails for BOM-prefixed XML / SVG written by
+     * text editors that auto-prepend a BOM (e.g. Notepad on Windows), and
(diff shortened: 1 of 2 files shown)


asolntsev (approved):

asolntsev:
Thank you @stanley-leong !

I've merged PR and published the result to Maven central repo as `10.2.1-SNAPSHOT`, you can try.

stanley-leong:
Thanks for the quick turnaround!

Tested 10.2.1-SNAPSHOT locally on a downstream Spring Boot service:

- Bumped `flying-saucer-core` / `flying-saucer-pdf` to `10.2.1-SNAPSHOT` and pulled from `https://central.sonatype.com/repository/maven-snapshots/`.
- Disabled the BOM-strip workaround I had wrapping the user-agent's `resolveAndOpenStream`.
- Regression test (BOM-prefixed SVG → drawable image XObject in the output PDF) passes with the workaround disabled against `10.2.1-SNAPSHOT`. The 3 existing PDF-render tests still pass too.

Looks good for `10.2.1`. Not pushing the SNAPSHOT to production on our side (will keep the local override until `10.2.1` is released), but happy to confirm the fix works as expected end-to-end on real-world BOM-prefixed SVGs.

Thanks again!

asolntsev:
@stanley-leong Cool!
I've released FS 10.2.1 - thank you for the collaboration!

Issue #589, opened 2025-10-21
Add HTML5 element support for section, article, header, footer


andreasrosdalw:
` Add simple rendering for <section>, <article>, <header>, <footer> - treat as block-level elements. `

Flying Saucer currently treats unknown HTML elements as inline by default. To improve HTML5 compatibility, we can treat modern sectioning elements as block-level elements, just like div. Apparently this should be pretty easy to do according to chatgpt.

NB! Don't forget to remove the implemented elements from `BoxBuilder.UNSUPPORTED_HTML5_TAGS` (added in https://github.com/flyingsaucerproject/flyingsaucer/pull/654).

PR #683, opened 2026-07-10
Make SharedContext non-final, so it can be extended

balrok:
Our use-case is to suppress some of the recent "unsupportedFeatures". For example #589 but also some CSS-features where we just accept that they are unsupported.

### DIFF
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/layout/SharedContext.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/layout/SharedContext.java
@@ -68,7 +68,7 @@ import static java.util.stream.Collectors.joining;
  *
  * @author empty
  */
-public final class SharedContext {
+public class SharedContext {
     private static final Set<String> PAGED_MEDIA_TYPES = Set.of("print", "projection", "embossed", "handheld", "tv");
 
     private TextRenderer textRenderer;


asolntsev:
@balrok I am afraid it's not a good idea to create such a coupling with FS core classes.
Do I correctly understand that you want to remove these warnings in logs?

```java
Encountered HTML5 elements which are not supported by FlyingSaucer: {}. Rendering may be incorrect.
Encountered CSS3 features not supported by FlyingSaucer: {}. Rendering may be incorrect.
```

Then I would ask first:
1. Don't you think these are reasonable warnings, and it's a good idea to replace such html tags in your html by some other tags supported by FS?
2. If you still want to ignore these warnings, isn't it better to just add a line to `log4j.properties` or `logback.xml` like this:
```properties
  org.slf4j.simpleLogger.log.org.xhtmlrenderer.pdf.ITextRenderer=error
```

If #2 seems good, we could write these warnings with a dedicated logger name, so you can ignore only logs from this logger.

asolntsev:
@balrok This is what I mean: https://github.com/flyingsaucerproject/flyingsaucer/pull/687

asolntsev:
I hope the initial problem was already solved by https://github.com/flyingsaucerproject/flyingsaucer/pull/687

PR #686, opened 2026-07-15
feat(pdf): support inline <svg> elements

asolntsev:
## Summary
- `<svg>...</svg>` written directly in the document body previously fell through to generic inline layout (its children rendered as unstyled boxes); only `` and CSS `background-image` SVGs worked.
- Give `svg` `display: inline-block` in the UA stylesheet and route its `width`/`height` presentational attributes through the same CSS conversion as `img`, so it becomes a proper replaced-element box.
- `ITextReplacedElementFactory` now serializes the inline `<svg>` subtree to XML and rasterizes it through the existing Batik-based `SvgImage`/`PNGTranscoder` pipeline (the same one used for ``).
- Extracted the width/height/viewBox sizing logic in `ITextUserAgent` into a reusable `getSvgSize(Element)` so both the URI-based and inline paths share it.

Note: as with any SVG (in a browser too), scaling an inline `<svg>` via `width`/`height` alone only works correctly if it also declares a `viewBox` — without one, the coordinate system stays 1 unit = 1px and shrinking just clips the artwork. `page-with-svg.html` was updated to include a `viewBox` for that reason.

## Test plan
- [x] `mvn -pl flying-saucer-core,flying-saucer-pdf -am test` passes
- [x] `SvgTest` renders `page-with-svg.html` to PDF and asserts the "SVG Star:" text is present
- [x] Manually rasterized the generated PDF to PNG and visually confirmed the inline SVG star renders correctly at its specified size, alongside the existing classpath-SVG/background-image cases

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/simple/extend/XhtmlCssOnlyNamespaceHandler.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/simple/extend/XhtmlCssOnlyNamespaceHandler.java
@@ -108,7 +108,8 @@ public class XhtmlCssOnlyNamespaceHandler extends NoNamespaceHandler {
                 style.append(e, "rowspan", "-fs-table-cell-rowspan: ");
                 break;
             }
-            case "img": {
+            case "img":
+            case "svg": {
                 style.appendWidth(e);
                 style.appendHeight(e);
                 break;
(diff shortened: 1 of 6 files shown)


asolntsev:
@coderabbitai review

PR #698, opened 2026-07-31
Fix server-relative URLs against synthetic inline stylesheet URIs

### LINKED ISSUE #695: Relative paths to fonts prefixed with "inline/" segment, if font-face directive exists as part of inline style block.
version org.xhtmlrenderer:flying-saucer-pdf:10.3.0

When inline style block parsed, CSSParser sets _uri to inline:1 for internal needs, but later, 
when binaries should be fetched, this _uri is used as prefix for the relative face font path. 
As a result, you write something like: 
```
@font-face {
...
    src: url('/assets/fonts/MyFont.ttf');
}

```
but resource path used by library is 'inline/assets/fonts/MyFont.ttf', which does not exists of course. 

Is this and issue or a feature? 
Because currently behaviour is pretty confusing.

kalayciburak:
## Summary
Inline `<style>` blocks are cached under synthetic opaque URIs such as `inline:1`. Those values are absolute URIs, so CSS `url(...)` resolution treated them as stylesheet bases and rewrote server-relative paths:

- `url('/assets/fonts/MyFont.ttf')` became `inline/assets/fonts/MyFont.ttf`

That breaks `@font-face` (and other resources) declared in inline CSS.

Server-relative path rewriting now requires a **hierarchical** absolute base (`URI.isAbsolute() && !URI.isOpaque()`), so `http(s):` / `file:` bases still work while opaque synthetic keys do not rewrite the path.

Fixes #695

## Test plan
- [x] RED: `CSSParserTest` reproduced `inline/assets/fonts/MyFont.ttf`
- [x] GREEN: `mvn -pl flying-saucer-core test -Dtest=CSSParserTest` — 19/19
- [x] `mvn -pl flying-saucer-core test` — 244/244
- [x] `mvn -pl flying-saucer-pdf -am test -Dtest=CssFontFaceTest,CSSParserTest` — GREEN

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resource URL handling for inline and synthetic stylesheets.
  * Prevented invalid rewriting of opaque or malformed URIs as filesystem paths.
  * Preserved correct absolute, relative, and server-relative paths, including font resources and `file:` URLs.

* **Tests**
  * Added coverage for URL resolution in inline stylesheets, style attributes, and `@font-face` declarations.

### DIFF
--- a/flying-saucer-core/src/main/java/org/xhtmlrenderer/css/parser/CSSParser.java
+++ b/flying-saucer-core/src/main/java/org/xhtmlrenderer/css/parser/CSSParser.java
@@ -2076,7 +2076,7 @@ public class CSSParser {
                 // Relative URIs are resolved relative to CSS file, not XHTML file
                 if (isRelativeURI(uriResult) && _uri != null) {
                     uriResult = getPartBeforeLastSlash(_uri) + uriResult;
-                } else if (isServerRelativeURI(uriResult) && _uri != null && isAbsoluteUri(_uri)) {
+                } else if (isServerRelativeURI(uriResult) && _uri != null && isHierarchicalAbsoluteUri(_uri)) {
                     uriResult = getPartBeforeFirstSlash(_uri) + uriResult;
                 }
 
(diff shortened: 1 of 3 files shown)


asolntsev:
@kalayciburak Can you add an integration test showing the problem?

See my comment in the original ticket: https://github.com/flyingsaucerproject/flyingsaucer/issues/695#issuecomment-5272848427

kalayciburak:
added an IT for the server-relative case (`url('/...')` in inline @font-face). relative paths like `../../fonts/...` still resolve against the html file so they never hit the `inline/` prefix. this one does, and the font actually gets embedded.

asolntsev (approved):

PR #682, opened 2026-07-07
fix(layout): honor box-sizing: border-box for display: table-cell

hamfer:
Prior to this fix, the box-sizing: border-box CSS property was silently
ignored for elements with display: table-cell, causing three distinct bugs:

1. WIDTH (TableCellBox.getOuterStyleWidth)
   getOuterStyleWidth() always added padding+border to the CSS width,
   treating it as content-box regardless of box-sizing. With border-box,
   the CSS width already includes padding+border, so they must not be
   added again.

   Before: width: 100px; padding-right: 10px → getWidth() = 110px ❌
   After:  width: 100px; padding-right: 10px → getWidth() = 100px ✅

2. HEIGHT (TableCellBox.getCSSHeight)
   getCSSHeight() always subtracted padding+border from the CSS height
   (XHTML 1.0 legacy quirk). When box-sizing: border-box was explicitly
   set, calcDimensions() subtracted them again, causing double-subtraction.

   Before (border-box): height: 80px; padding: 10px → getHeight() = 60px ❌
   After  (border-box): height: 80px; padding: 10px → getHeight() = 80px ✅
   After  (content-box): legacy XHTML 1.0 quirk preserved          → 80px ✅

3. MIN/MAX-WIDTH (BlockBox.applyCSSMinMaxWidth + TableCellBox.setLayoutWidth)
   Two related issues:
   a) applyCSSMinMaxWidth() clamped raw content width against min/max-width
      values without accounting for padding+border when border-box was set.
   b) setLayoutWidth() overwrote the content width already constrained by
      applyCSSMinMaxWidth() in calcDimensions(), so constraints were lost.

   Before: max-width: 80px; padding: 20px → getWidth() = 200px ❌
   After:  max-width: 80px; padding: 20px → getWidth() =  80px ✅

Root cause: box-sizing is a CSS3 property. Flying Saucer's infrastructure
already parsed and stored it (CSSName.BOX_SIZING, IdentValue.BORDER_BOX,
CalculatedStyle.isBorderBox()), and BlockBox.calcDimensions() already used
isBorderBox() for normal block elements. The table-cell-specific code paths
in TableCellBox and the min/max-width logic were not updated accordingly.

Note: table-layout: auto is unaffected by the getOuterStyleWidth fix because
AutoTableLayout uses calcMinMaxWidth() for initial column sizing, which goes
through the standard BlockBox.calcDimensions() path that was already correct.
Use table-layout: fixed to ensure border-box is fully respected on table cells.

Files changed:
[...]

PR #702, opened 2026-08-07
fix(layout): skip applyCSSMinMaxWidth in setLayoutWidth for content-box table cells

hamfer:
PR #682 added a call to `applyCSSMinMaxWidth(c)` at the end of
`TableCellBox.setLayoutWidth()` to honour `box-sizing: border-box` on table
cells. However, the call was added **unconditionally**, which means it also
runs for `content-box` cells (the CSS default).

In `table-layout: fixed`, the `width` parameter passed into `setLayoutWidth()`
is already the authoritative outer width allocated by the fixed table-layout
algorithm — summed across any `colspan` spans from `<colgroup>`/`<col>`.
Calling `applyCSSMinMaxWidth(c)` afterwards allows CSS `max-width`/`min-width`
declared on the `<td>` to **silently override that allocation**, which breaks
text wrapping inside the cell.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fixed-layout table rendering for column groups and cells spanning multiple columns.
  * Corrected width calculations while preserving allocated column widths.
  * Improved minimum and maximum width handling for border-box and content-box cells.
  * Fixed text wrapping and line placement for content with maximum-width styling.

* **Tests**
  * Added PDF regression coverage for fixed- and auto-layout tables, column allocation, colspan sizing, and expected text wrapping.


asolntsev:
@hamfer My AI gave such a feedback.


## Code review

The core issue: the fix gates `applyCSSMinMaxWidth` on `box-sizing: border-box` (`getStyle().isBorderBox()`), but the distinction that actually matters here is `table-layout: fixed` vs `auto`. The codebase already has the right predicate for this — `TableCellBox.isFixedWidthAdvisoryOnly()` (line 501), which checks `table-layout: auto`.

As written:

1. **Regresses `content-box` cells in `table-layout: auto` tables.** `TableSectionBox.setCellWidths()` (line 215) calls `cell.setLayoutWidth()` unconditionally for every cell, regardless of `AutoTableLayout` vs `FixedTableLayout`. A plain `table-layout: auto` table with a content-box `<td style="max-width:150px">` now has `max-width` silently ignored — this worked before this PR.
2. **Doesn't actually fix the bug for `border-box` cells.** In `table-layout: fixed`, a `border-box` colspan `<td>` with `max-width` still gets `applyCSSMinMaxWidth` applied unconditionally, so the exact regression from #682 (max-width overriding the col-allocated width) still reproduces whenever the cell happens to be `border-box`. No test covers this case.

Suggestion: gate on `isFixedWidthAdvisoryOnly()` (or equivalently, check `table-layout: fixed`) instead of `isBorderBox()`.

Two smaller nits in `TableFixedLayoutColgroupColspanWrapTest.java`:

- Several tests manually unwrap `pdf.text` and assert with plain AssertJ (e.g. line 109) instead of using `PdfAssert.containsExactText(...)`, which already exists in `pdf-test` and does the same check.
- `etaAndThetaAreOnSameLineAtCorrectCellWidth` infers which cell width was used from hand-computed character-width/line-wrap arithmetic (`~5pt/char`, `~74 chars/line`). This is fragile to font-metric or OpenPDF version changes unrelated to the bug being guarded against.

